PR #8, opened 2017-12-05
Develop

TAlonglong:
Small adaptions from MET Norway to make thing run smothly for us.

Should not affect other users.


pnuu:
Looks good to me. Merging.

PR #13, opened 2018-05-29
Add Python 3 support

### LINKED ISSUE #12: Python 3 support
Some of the modules/scripts have Python 2 specific syntax, imports and/or dependencies. These should be converted so that everything works also with Python 3.

pnuu:
This PR adds Python 3 support to pytroll-collector scripts/modules.

- [x] handle urlparse <-> urllib.parse
- [x] handle ConfigParser <-> configparser
- [x] print <-> print()
- [x] remove mpop dependencies

Fixes #12


pnuu:
Now everything should work with both Python 2 and Python 3.

I had to remove blending between the images in the global mosaic creator, as it didn't work. And when it worked with Py2, it created suboptimal results. I'll have another PR for it at a later time (I already started something for it).

PR #26, opened 2019-01-08
add on_moved process to watchdog trigger.

yufeizhu600:
In the case that a file has been moved to the target directory, python watchdog may only generated a 'moved' event; since there is no 'on_moved' method of the watchdogtrigger, the file will not trigger the processing even the file name matches the pattern. 

This PR is to add on_moved event process to the watchdog trigger to deal with above case.


pnuu:
Looks good to me. The code is almost identical to `on_created()`, so maybe they could call a common method and pass the appropriate path from the event to it?

PR #28, opened 2019-02-01
Bugfix segment gatherer empty fileset

pnuu:
Segment gatherer added an filename with empty segment/channel to the list of "wanted files" if e.g. "critical files" were empty in an INI-based config. This PR fixes that, and additionally fixes unittests that failed due to earlier similar changes to YAML-based config.


pnuu:
Tested this in production, and works as intended.

PR #40, opened 2019-09-17
PyToCh Gatherer: Add hour filter pattern

tecnavia-dev:
Some new features added:
Gatherer
- Add the possibility to filter raw image by hour to be able to process data in different interval range respect the raw data. To configure the new feature add the following lines:
#hour_pattern: "StartHH:mm EndHH:mm DeltaHH:mm"
hour_pattern: "00:00 23:59 00:30"  #filter every 30 min
hour_pattern: "06:00 18:00 01:00"  #filter of 06:00 06:30 ... until 18:00


mraspaud:
Thanks for submitting all the changes. As these provide multiple functionality, would it be possible for you to make one PR per functionality instead ?  One for gatherer and two for trollstalker. It will be easier to discuss and improve on each one that way.

Regarding the `launch.json` file, I think it's best not to integrate it in this package:
- everyone uses different editors, which makes it unpractical if a config file for every editor was to be provided
- the file you provide seem to have path specific to you system, so it might not be usable by other.

mraspaud:
Also, could you change the title of this PR to make it more descriptive ? Right now it doesn't say anything on the content of the PR.

mraspaud:
Can I close this PR ? It looks like other PRs are going to replace this one, right ?

tecnavia-dev:
Yes and also #33 of Vincenzo

mraspaud:
Thanks

PR #41, opened 2019-09-18
Add filter hour pattern to segment gatherer

tecnavia-dev:
Some new features added:
Gatherer

- Add the possibility to filter raw image by hour to be able to process data in different interval range respect the raw data. To configure the new feature add the following lines:
#hour_pattern: "StartHH:mm EndHH:mm DeltaHH:mm"
hour_pattern: "00:00 23:59 00:30" #filter every 30 min
hour_pattern: "06:00 18:00 01:00" #filter of 06:00 06:30 ... until 18:00


mraspaud:
Thank for the pull request. Can you fix the linter issues and provide an example config for the new feature ?

examples/segment_gatherer_msg_hour.yaml_template
```diff
@@ -0,0 +1,42 @@
+# Filename patterns that are collected
+patterns:
+  msg:
+    pattern:
+      "H-000-{orig_platform_name:4s}__-{orig_platform_name:4s}________-{channel_name:_<9s}-{segment:_<9s}-{start_time:%Y%m%d%H%M}-__"
+    # Process images hourly
+    hour_pattern: "00:00 23:59 01:00"
```
mraspaud: This is not very obvious what each value is, could we use instead something like:
```yaml
hour_pattern:
    start_time: "00:00"
    end_time: "23:59"
    time_delta: "01:00"
```

pytroll_collectors/segments.py
```diff
@@ -76,6 +76,30 @@ def __init__(self, config):
         self.logger = logging.getLogger("segment_gatherer")
         self._loop = False
 
+        # Convert check time into int minutes variables
+        for key in self._patterns:
+            if "hour_pattern" in self._patterns[key]:
+                checkTime = self._patterns[key]["hour_pattern"]
+                # Ex. checkTime =  "06:00 18:00 00:30"
+                startTime, endTime, deltaTime = checkTime.split(' ')
+                if startTime != '' and endTime != '' and deltaTime != '':
```
mraspaud: the suggestion above would allow you to get the values directly instead of having to split them. Also, could you make start and end times optional ? (defaulting to 00:00 and 23:59 respectively)

pytroll_collectors/segments.py
```diff
@@ -489,6 +524,23 @@ def _find_time_slot(self, time_obj):
...
+        if checkTime["midnight"] and rawT < checkTime["start"]:
+            rawT += 24*60
+
+        # Check start end time
+        if rawT >= checkTime["start"] and rawT <= checkTime["end"]:
+            # Raw time in range, check interval
+            if ((rawT - checkTime["start"]) % checkTime["delta"]) == 0:
+                toret = 1
+
+        return toret
```
mraspaud: Python uses boolean types, like True and False. Please use that instead of 0 and 1.

examples/segment_gatherer_msg_hour.yaml_template
tecnavia-dev: Yes, integrated.
But if I will produce for example the IR all the day and Visible only during the day, where I can configure this?
The gatherer sent using a topic and where, I can separate the processing ?

mraspaud: In trollflow2 for example. There you can define for each composite a min or max sunlight coverage for example.

pytroll_collectors/segments.py
```diff
@@ -75,6 +75,39 @@ def __init__(self, config):
...
+                time_conf = self._patterns[key]["hour_pattern"]
+                start_time_str = "00:00"
+                end_time_str = "23:59"
+                delta_time_str = "00:01"
+                if "start_time" in time_conf:
+                    start_time_str = time_conf["start_time"]
+                if "end_time" in time_conf:
+                    end_time_str = time_conf["end_time"]
+                if "delta_time" in time_conf:
+                    delta_time_str = time_conf["delta_time"]
```
mraspaud: instead of
```python
start_time_str = "00:00"
if "start_time" in time_conf:
  start_time_str = time_conf["start_time"]
```
you can write:
```python
start_time_str = time_conf.get("start_time", "00:00")
```

pytroll_collectors/segments.py
```diff
@@ -75,6 +75,39 @@ def __init__(self, config):
...
+
+                start_h, start_m = start_time_str.split(':')
+                end_h, end_m = end_time_str.split(':')
+                delta_h, delta_m = delta_time_str.split(':')
+                hour = {}
```
mraspaud: I think `hour` is a bit missleading as a name for this variable. How about `interval` ?

pytroll_collectors/segments.py
```diff
@@ -404,6 +437,17 @@ def process(self, msg):
...
 
+        # Check if time of the raw is in scheduled range
+        if "_hour_pattern" in self._patterns[key]:
+            schedule_ok = self.check_schedule_time(
+                self._patterns[key]["_hour_pattern"],
+                metadata["start_time"].hour,
+                metadata["start_time"].minute)
+            if not schedule_ok:
+                self.logger.info("Hour pattern '%s' skip: %s",
+                                 key, msg.data["uid"])
```
mraspaud: could you maybe also print the time step here that is being skipped ?

pytroll_collectors/segments.py
```diff
@@ -491,6 +535,23 @@ def _find_time_slot(self, time_obj):
 
         return str(time_obj)
 
+    def check_schedule_time(self, check_time, raw_hour, raw_minute):
+        # Check if raw time is inside configured interval
+        toret = False
```
mraspaud: It took me some time to understand what `toret` meant, can you rename this to `to_ret` or even something more specific like eg `time_ok` ?

pytroll_collectors/segments.py
```diff
@@ -491,6 +535,23 @@ def _find_time_slot(self, time_obj):
 
         return str(time_obj)
 
+    def check_schedule_time(self, check_time, raw_hour, raw_minute):
```
mraspaud: I think it would be better to pass a python time object instead of hour+minute. That limits the number of arguments to the function, and won't break backwards compatibility if we decide to add eg seconds in the future.

pytroll_collectors/segments.py
```diff
@@ -491,6 +535,23 @@ def _find_time_slot(self, time_obj):
...
+        # Check if raw time is inside configured interval
```
mraspaud: A docstring should start and finish with """ (better than #)
Also you need a . at the end of your sentence :)

pytroll_collectors/segments.py
```diff
@@ -75,6 +75,39 @@ def __init__(self, config):
         self.logger = logging.getLogger("segment_gatherer")
         self._loop = False
 
+        # Convert check time into int minutes variables
+        for key in self._patterns:
+            if "hour_pattern" in self._patterns[key]:
```
mraspaud: from what I understand, this patch operates on the start time of the data only, right ? so could this be renamed to reflect this, for example `start_time_pattern` ?

tecnavia-dev: Maybe if we call the filter "slot time" is it more generic for all satellites ?

mraspaud: "Slot time" is more generic indeed, but then it's the same problem as "hour_pattern", which is that we don't know from the name if it operates on `start_time`, `nominal_time`, `processing_time` or `end_time` for example...

tecnavia-dev: For the hour_pattern, really in the code we test the metadata start_time received from the trollstalker, start_time_pattern is the most clear value to be used to compare.

mraspaud:
@tecnavia-dev any updates on this PR ?

pytroll_collectors/segments.py
```diff
@@ -508,6 +548,24 @@ def _find_time_slot(self, time_obj):
 
         return str(time_obj)
 
+    def check_schedule_time(self, check_time, raw_start_time):
+        """Check if raw time is inside configured interval."""
+
```
mraspaud:
```suggestion

```

pytroll_collectors/tests/test_segments.py
```diff
@@ -376,6 +382,20 @@ def test_ini_to_dict(self):
         self.assertTrue('is_critical_set' in config['patterns']['msg'])
         self.assertTrue('variable_tags' in config['patterns']['msg'])
 
+
```
mraspaud:
```suggestion
```

PR #45, opened 2019-10-11
Add commandline options for publisher port and publisher nameservers

pnuu:
This PR adds to command-line arguments for `gatherer.py`:
 - `-n`, `--nameservers`: nameservers to which publisher connects. Defaults to `localhost` if not defined.
 - `-p`, `--publish_port`: port in which the messages are published. Defaults to random port if not defined.

### DIFF
--- a/bin/gatherer.py
+++ b/bin/gatherer.py
@@ -132,6 +132,11 @@ def arg_parse():
     parser.add_argument("-c", "--config-item",
                         help="config item to use (all by default). Can be specified multiply times",
                         action="append")
+    parser.add_argument("-p", "--publish-port",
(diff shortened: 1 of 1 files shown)


bin/gatherer.py
```diff
@@ -247,9 +252,20 @@ def main():
...
 
+    if opts.publish_port:
+        publish_port = int(opts.publish_port)
+    else:
+        publish_port = 0
+
+    if opts.nameservers:
+        publisher_nameservers = opts.nameservers.split(',')
+    else:
+        publisher_nameservers = None
```
mraspaud: Can we use the argparse mechanism to do this instead ? (provide the type, default value, etc)

PR #46, opened 2019-10-14
Added unique service name where possible. 

TAlonglong:
Added unique service name to the cat, gatherer and trollstalker.

The segment_gathere uses the same publisher for all configured segment gatherings so this can not be used for this.

This way each instanse running av a unique service name and can be identified from the nameserver

### DIFF
--- a/bin/cat.py
+++ b/bin/cat.py
@@ -216,7 +216,7 @@ if __name__ == '__main__':
         config['service'] = ''
 
     try:
-        with Publish("cat") as pub:
+        with Publish("cat_" + opts.config_item) as pub:
             with Subscribe(config['service'], config["topic"], True) as sub:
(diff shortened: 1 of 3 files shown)


TAlonglong:
@mraspaud or @pnuu. What do you think about this?

mraspaud:
I don't think we (SMHI) are using the service in any way atm. And in theory I think this is a good idea.

TAlonglong:
I dont think so @pnuu. If you dont use service then this has no effect. It is just the name the publisher is registered by in the nameserver. But if you use the service, then you have to update the config.
But this will reduce the number of services with the same name and therefor reduce the number of connections.

pnuu:
Ok, then I'm fine with this :+1:

PR #47, opened 2019-11-18
Make it possible to group segments for full minutes

pnuu:
This PR makes it possible to group segments for given full minutes.

The feature makes it possible to collect e.g. Himawari-8 segments that otherwise would require 6 separate collectors (one for each 10-minute slot), and then would have incorrect time attached for slots not starting at full hour. To configure (see full example in `examples/segment_gatherer.ini_template`, `[himawari-8]` section):

```ini
pattern = IMG_{platform_name:4s}{channel_name:3s}_{start_time:%Y%m%d%H%M}_{segment:0>3s}
...
group_by_minutes = 10
keep_parsed_keys = start_time
```

This means that the segments are group for every full 10 minute period. This period can be anything from `1` (group for every full minute) onwards. Values equal or greater than `60` will group the data to full hours.

The last option means that the `start_time` parsed from the filename is not overwritten by the time given in the incoming message. Any other tags in the pattern can be given in same way.

### DIFF
--- a/pytroll_collectors/tests/data/segments.ini
+++ b/pytroll_collectors/tests/data/segments.ini
@@ -31,4 +31,4 @@ time_name = start_time
 variable_tags = start_minute
 add_minutes = 10
 add_minutes_multiplier = start_ten_minutes
-use_message_start_time = False
+keep_parsed_keys = start_time
--- a/pytroll_collectors/tests/data/segments.ini
+++ b/pytroll_collectors/tests/data/segments.ini
@@ -20,7 +20,7 @@ time_name = start_time
 variable_tags = end_time,end_frac_sec,creation_time
 
 [himawari-8]
-pattern = IMG_DK01{channel_name:3s}_{start_time:%Y%m%d%H}{start_ten_minutes:1d}{start_minute:1d}_{segment:0>3s}
+pattern = IMG_DK01{channel_name:3s}_{start_time:%Y%m%d%H%M}_{segment:0>3s}
 critical_files = B01:001-010,B02:001-010,B04:001-010,B05:001-010,B09:001-010,B10:001-010,B11:001-010,B14:001-010,B16:001-010,IR1:001-010,IR2:001-010,IR3:001-010,IR4:001-010,VIS:001-010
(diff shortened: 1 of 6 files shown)


pnuu:
Tested this with real-time data from Eumetcast. Rather surprisingly works :grin:

TAlonglong:
As far as I can see this looks good.

pnuu:
Thanks to @mraspaud for pointing out that the previous solution was pretty convoluted. This should be much easier for the user to understand and configure.

TAlonglong:
Nice @pnuu !

Could you add some explanation at least in the ini file of the `group_by_minutes` variable?

pnuu:
Thanks @TAlonglong! I had also forgotten to adjust the file pattern for the Himawari-8 example config.

pytroll_collectors/tests/data/segments.ini
```diff
@@ -18,3 +18,16 @@ publish_topic = /pub/foo/bar
...
+[himawari-8]
+pattern = IMG_DK01{channel_name:3s}_{start_time:%Y%m%d%H%M}_{segment:0>3s}
+critical_files = B01:001-010,B02:001-010,B04:001-010,B05:001-010,B09:001-010,B10:001-010,B11:001-010,B14:001-010,B16:001-010,IR1:001-010,IR2:001-010,IR3:001-010,IR4:001-010,VIS:001-010
+wanted_files = B01:001-010,B02:001-010,B04:001-010,B05:001-010,B06:001-010,B09:001-010,B10:001-010,B11:001-010,B12:001-010,B14:001-010,B16:001-010,IR1:001-010,IR2:001-010,IR3:001-010,IR4:001-010,VIS:001-010
+all_files = B01:001-010,B02:001-010,B04:001-010,B05:001-010,B06:001-010,B09:001-010,B10:001-010,B11:001-010,B12:001-010,B14:001-010,B16:001-010,IR1:001-010,IR2:001-010,IR3:001-010,IR4:001-010,VIS:001-010
+topics = /himawari/H8/L1
+publish_topic = /himawari/H8/segment
+timeliness = 1200
+time_name = start_time
+variable_tags = start_minute
```
mraspaud: This isn't needed anymore, right ?

pnuu: Thanks! Removed `variable_tags` from both Himwari configs (example template and tests).

PR #48, opened 2019-11-22
Use versioneer for versioning

pnuu:
This PR changes versioning to use `versioneer`. Also, instructions for releasing are added.


pnuu:
Apparently versioneer isn't a thing anymore, so closing as obsolete.

PR #51, opened 2019-11-29
Add providing server, nameservers and services as parameters to gatherer and cat

TAlonglong:
Some additions to the cat and segments

Add nameservers config for cat
Add possible providing server and servers to segments

### DIFF
--- a/bin/cat.py
+++ b/bin/cat.py
@@ -226,7 +226,6 @@ if __name__ == '__main__':
                 for msg in sub.recv(2):
                     if msg is None:
                         continue
-                    LOG.info("msg type: %s", msg.type)
                     if msg.type == "collection":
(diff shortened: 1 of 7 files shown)


TAlonglong:
I need to have the possibility to add nameservers for the cat module. My suggestion is to add this in the config. This must be added for each cat configuration. In gatherer the nameservers is given on the command line, but there the gatherer uses one publisher for all gatherings in the config file so that make sense.

TAlonglong:
Thanks for the title change.

TAlonglong:
After reviewing myself I saw there was a mix if service and services. Only use the last one.

Added some examples to the features added

TAlonglong:
This might need a new review @mraspaud  and @pnuu

PR #53, opened 2019-12-17
Feature limit gatherer by schedules

### LINKED ISSUE #52: Add option to adjust start and/or end time after pass calculation is made in gatherer
I'm thinking of how I can add an option adjust the start and or end time of the gatherer.

When a file is received the possible files are predicted by the gatherer for the area given to the gatherer.

For EARS data this is mainly ok, but for MERSI2 data the coverage is (for the moment) bad for the EARS area due to low prioritation and overlapp with Suomi NPP passes.

The ears mersi prediction from Eumetsat can be seen( or downloaded) here:
https://uns.eumetsat.int/ see EARS pass prediction and scroll down to the files with mersi in,

or eg.:

https://uns.eumetsat.int/downloads/ears/ears_mersi_pass_prediction_19-11-29.txt

Example of Passes for can be seen [here](https://drive.google.com/file/d/1KQohM4Edus1JWh6gE8ht4OwU3wYGZn9v/view?usp=sharing)

I think something similar is done with AQUA TERRA for antenna scheduling where the DB is scheduled to be off.

So my suggestion is to let the gatherer to the pass calculation as done today. Then download the EUM pass file ( somehow configured in the config file), and then adjust the start and/or end time of the calculated pass to avoid the gatherer waiting for segments that never will arrive.

TAlonglong:
Download ears schedules from EUM, parse and limit `planned_granule_times`.

### DIFF
--- a/bin/gatherer.py
+++ b/bin/gatherer.py
@@ -156,7 +156,19 @@ def setup(decoder):
             duration = timedelta(seconds=CONFIG.getfloat(section, "duration"))
         except NoOptionError:
             duration = None
-        collectors = [region_collector.RegionCollector(region, timeliness, duration)
+        # Parse schedule cut if configured. Mainly for EARS data.
+        try:
+            schedule_cut = CONFIG.get(section, 'schedule_cut')
+        except NoOptionError:
+            schedule_cut = None
+        # If you want to provide your own method to provide the schedule cut data
+        try:
+            schedule_cut_method = CONFIG.get(section, 'schedule_cut_method')
+        except NoOptionError:
+            schedule_cut_method = None
+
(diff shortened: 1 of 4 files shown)


TAlonglong:
Thanks. 

For the test I think it will be some time before I can do this.

adybbroe:
Any progress recently here?
@TAlonglong can you have a look ?

pnuu:
As (briefly) discussed on Slack, I'll re-implement this feature on top of the refactoring being done on #85 . Rebasing/merging will make things unnecessarily complicated, so starting a new PR will be the easier route. I'll leave this open so it's easier to find, but convert to a draft.

pnuu:
Superseded by #90

PR #57, opened 2020-02-13
Add more config options to cat.py

pnuu:
This PR adds the following config file options for `cat.py`:

- `publish_port` - port number where messages are published. Optional, defaults to random port.
- `subscriber_nameserver` - connect subscriber to this nameserver. Optional, defaults to localhost.
- `subscriber_addresses` - connect directly to listed servers and ports. Optional, not used by default.


bin/cat.py
```diff
@@ -220,21 +227,39 @@ def process_message(msg, config):
     if 'nameservers' in config:
         nameservers = config['nameservers'].split(',')
 
+    publish_port = 0
+    if 'publish_port' in config:
+        publish_port = int(config['publish_port'])
+
+    sub_nameserver = 'localhost'
+    if 'subscriber_nameserver' in config:
+        sub_nameserver = config['subscriber_nameserver']
```
mraspaud, line 236: The default before this change was [], right ?

bin/cat.py
```diff
@@ -220,21 +227,39 @@ def process_message(msg, config):
...
+
+    sub_addresses = None
+    if 'subscireber_addresses' in config:
```
mraspaud, line 239:
```suggestion
    if 'subscriber_addresses' in config:
```

bin/cat.py
mraspaud: my bad, it was `localhost`

PR #58, opened 2020-02-17
Discard messages without correct type in gatherer

pnuu:
This PR discards messages in `gatherer.py` if the message type is incorrect. The allowed types are `'file'`, `'dataset'` and `'collection'`. This is needed because trollmoves client is publishing a message with type `'ack'`, and these should not be caught by gatherer, as they are inteded only to notify other trollmoves clients that the message is being handled.


mraspaud:
No test ?

PR #60, opened 2020-03-19
group_by_minutes and keep_parsed_keys per patterns

TAlonglong:
If you have more than one patterns you might need to set the group_by_minute and keep_parsed_keys in the config different for each patterns.

eg: (yaml config)
```
patterns:
  msg-iodc:
    .... 
  himawari:
    ...
    group_by_minutes: 10
    keep_parsed_keys: start_time
```

Without this PR it is only possible to give these for all patterns.


TAlonglong:
Pleas give this a look when you can spare some minutes @mraspaud and @pnuu

mraspaud:
Thanks for fixing the minutes grouping with the hour patterns feature. Could you add a test ? This will probably help me understand the modifications better.

TAlonglong:
Hm ok. I will see what I can do

But is the test, at least for segments ( test_segments.py) run at all? If I try at the command line it fails.

https://travis-ci.org/github/pytroll/pytroll-collectors/jobs/664396828?utm_medium=notification&utm_source=github_status

Can you please have a look when you have time?

mraspaud:
I'll have a quicklook now

TAlonglong:
Tada

mraspaud:
Looks good @TAlonglong, thanks for rooting out the testing/coverage problem.
@tecnavia-dev it would be great if you could have a look at this! I'll merge this tomorrow if I don't hear from you.

PR #62, opened 2020-03-24
publish service name needs to be equal each time 

TAlonglong:
publish service name needs to be equal each time  even if patterns is an unordered dict


TAlonglong:
Could you @pnuu or @mraspaud have a look?

TAlonglong:
Fixed the conflict.

PR #64, opened 2020-09-15
Add the possibility to handle collections and datasets in input messages.

mraspaud:
Includes refactoring.


pnuu (approved):
A huge effort, nice job! Can't spot anything "wrong" here. If there's something that doesn't work on existing production chains, lets handle them in new PRs and add more tests to catch them.

mraspaud:
Thanks for the review. I will fix the codefactor issues and do a bit of live testing before I merge.

PR #81, opened 2021-04-13
Add Github Action to run unit tests

pnuu:
This PR adds a Github Action that runs the unit tests. Non-functional Travis CI integration is remvoed.


pnuu:
There are plenty of cases where `Thread.__init__(self)` is used, but they should be something like `super(NameOfThisClass, self).__init__()`. As the test coverage is pretty poor and I have no intention on adding tests on this PR, I just fixed the one that caused the tests to fail.

mraspaud (approved):
LGTM! Thanks for restoring CI :)

PR #84, opened 2021-04-21
Add a way to populate segment gatherer with pre-existing files

pnuu:
This PR adds a config option that makes it possible to populate the segment gatherer with pre-existing files. The check is triggered only for the first received message after `segment_gatherer.py` is started. Only `"file"` messages and local filesystem are supported.

As use case, consider the segmented MSG SEVIRI data from Eumetcast. If a restart is required due to configuration change or upgrade, currently the restart needs to be done either during the short gap between two timeslots, re-send the messages from new files, or potentially lose all files that arrived before the restart, or even the full set if prologue or epilogue files are missed.

With this PR, the first incoming message triggers (if `check_existing_files_after_start` is set to `true` in config) a check for all the files in the same directory where the `'uri'` in the message points to. The matching files will be added to the newly created slot.

### DIFF
--- a/examples/segment_gatherer_msg.yaml_template
+++ b/examples/segment_gatherer_msg.yaml_template
@@ -38,3 +38,7 @@ posttroll:
   # List of addresses for incoming messages. Default: [] (use multicast)
   # format: 123.456.78.9:12345
   # addresses: null
+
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
LGTM

PR #88, opened 2021-05-19
Remove 'six' as requirement

### LINKED ISSUE #87: Remove usage of six
As we don't support Python 2.7 anymore, the usage of the `six` compatibility library can be removed.

pnuu:
This PR removes the now-obsolete usage of `six` compatibility library, closes #87 .

There were also two "hacks" to make Python 2 work, which were removed. And the headers in each file were harmonized.


mraspaud (approved):
LGTM, thanks for taking the time!

PR #90, opened 2021-05-21
Limit geographic gathering using overpass schedules

### LINKED ISSUE #52: Add option to adjust start and/or end time after pass calculation is made in gatherer

pnuu:
This is a rework of #53 after the restructuring of geographic gathering.

Closes #52


pnuu:
Hmm, GitHub seems to allow only Martin, Lars, Adam and Dave. I'll see if I can find a way to add you.

pnuu:
@TAlonglong ok, I added you as a maintainer for the repository and then you were available as a reviewer.

TAlonglong (commented):
Not sure how we should do this. Most of this is what was added in a previous PR I think so I feel responsible for this.
There are some flake8 corrections needed and some lines that er not tested.
I guess I can not contribute to this PR?

pnuu:
You should now be able to add commits, as you are maintainer of the repository.

TAlonglong:
Do you mind if I try to clean up a bit?

pnuu:
Please do :grin:

TAlonglong:
Hm, I can't figure out how to do this. The branch is in your fork. And I guess I need to contribute to that for this to be added in this PR.

TAlonglong:
OK @pnuu, Some of the test did not actually test much, so I added some checks. With this work I saw there was actually to readlines call after opening a cached file. So I fixed this also.

The codefactor is some temporary directory and use of urlopen. And, I guess, a big job refactor the region_collector. I will leave that for a later task? What do you think?

pnuu:
If you think this is ready, I'm ok for merging. Does @mraspaud have any comments?

pytroll_collectors/harvest_schedules.py
```diff
@@ -0,0 +1,127 @@
...
+    else:
+        logger.error("sensor not given in params in granule_metadata. Can not continue.")
+        return (None, None)
+    eum_url = EUM_BASE_URL + pass_list_file
+    save_file = os.path.join(save_basename, pass_list_file)
+    logger.debug("Pass list save file, %s", save_file)
+    return eum_url, save_file
+
+
+def harvest_schedules(params, save_basename='/tmp', eum_base_url=EUM_BASE_URL):
```
mraspaud, line 102:
```suggestion
def harvest_schedules(params, save_basename=None, eum_base_url=EUM_BASE_URL):
    if save_basename is None:
        save_basename = tempfile.gettempdir()
```

pytroll_collectors/harvest_schedules.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.debug("params: %s", params)
```
mraspaud, line 103:
```suggestion
    logger.debug("harvest_schedules params: %s", params)
```

pytroll_collectors/tests/test_harvest_schedules.py
```diff
@@ -0,0 +1,257 @@
...
+            logs = [rec.message for rec in self._caplog.records]
+            self.assertIsNone(min_time)
+            self.assertIsNone(max_time)
+            self.assertIn("Reading from cached files", str(logs))
+
+    @mock.patch('pytroll_collectors.harvest_schedules.urlopen')
+    def test_harvest_schedule_HTTPError(self, mock_harvest_schedules):
+        from urllib.error import HTTPError
+        mock_harvest_schedules.side_effect = HTTPError('This failed', 0, '', '', None)
+        # self.cloud_image.fetch()
```
mraspaud, line 188: I this useful or should it be removed?

pytroll_collectors/tests/test_harvest_schedules.py
```diff
@@ -0,0 +1,257 @@
...
+              '/data/pytroll/VIIRS-EARS/SVMC_npp_d20191216_t1344091_e1345333_b42154_c20191216135112000242_eum_ops.h5',
+              'uid': 'SVMC_npp_d20191216_t1344091_e1345333_b42154_c20191216135112000242_eum_ops.h5'}],
+            'start_decimal': 1, 'proctime': '20191216135112000242', 'antenna': 'ears',
+            'data_processing_level': '1B', 'tle_platform_name': 'SUOMI NPP', 'platform_name': 'npp',
+            'end_time': datetime.datetime(2019, 12, 16, 13, 45, 33), 'type': 'HDF5',
+            'collection_area_id': 'eurol'}
+
+        planned_granule_times = set([datetime.datetime(2019, 12, 13, 13, 19),
+                                     datetime.datetime(2019, 12, 13, 13, 38),
+                                     datetime.datetime(2019, 12, 13, 13, 27)])
```
mraspaud, line 113: This info is used in multiple tests, could they be moved to `setUp` instead?

pytroll_collectors/tests/test_harvest_schedules.py
```diff
@@ -0,0 +1,257 @@
...
+    """Test suite."""
+    loader = unittest.TestLoader()
+    mysuite = unittest.TestSuite()
+    mysuite.addTest(loader.loadTestsFromTestCase(TestHarvestSchedules))
+
+    return mysuite
+
+
+if __name__ == "__main__":
+    unittest.TextTestRunner(verbosity=2).run(suite())
```
mraspaud, line 257: Since we are using pytest, this is not needed for automated testing at least.

pytroll_collectors/region_collector.py
```diff
@@ -237,6 +243,51 @@ def _predict(self, granule_metadata, step):
...
+        Any other source can be implemented in a method passed in the configuration.
+        The method will be pasted a dict (see params below) and the method must return
+        two datetimes, minimum and maximum allowed time. The self.planned_granule_times
+        will then be modified accordingly.
+        """
+        if self.schedule_cut:
+            method_file_name = "pytroll_collectors.harvest_schedules"
+            name = "harvest_schedules"
+            if self.schedule_cut_method:
+                logger.debug("Use custom schedule cut method provided i config file...")
```
mraspaud, line 263:
```suggestion
                logger.debug("Use custom schedule cut method provided in config file...")
```

pytroll_collectors/harvest_schedules.py
```diff
@@ -0,0 +1,127 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Harvest schedules. Download schedules from EUM and parse to limit gatherer"""
```
mraspaud, line 24: If I understand correctly, this module parses EUM schedules. Could the name of the module and/or functions reflect this?

mraspaud (changes requested):
Thanks a lot for putting this together! Very nice with added flexibility of passing custom functions in the config file.

I have a few inline comments, and then I don't see any testing for the custom harvesting function passing?

pytroll_collectors/tests/test_harvest_schedules.py
TAlonglong: Its needed to check that a `HTTPError` gives the expected return if this is raised.

pytroll_collectors/tests/test_harvest_schedules.py
TAlonglong: OK

pytroll_collectors/tests/test_harvest_schedules.py
TAlonglong: OK, I think I have copy/pasted some unittest framwork and mixed up with pytest.

Will try to clean up

TAlonglong:
Ah, thats a good point. I will look into this.

pytroll_collectors/harvest_schedules.py
TAlonglong: Hm, good point.

TAlonglong: I think its enough to make the module reflect this. Then it will be easy to add new modules downloading and parsing other schedules using the same function name. And also for the user to provide custom modules.

pytroll_collectors/harvest_schedules.py
mraspaud: Maybe you need to import tempfile here also

pytroll_collectors/tests/test_harvest_schedules.py
mraspaud: I mean the comment out line :)

pytroll_collectors/harvest_schedules.py
mraspaud: Ah, I see now you did this already, forget it :)

pytroll_collectors/tests/test_harvest_schedules.py
```diff
@@ -95,49 +110,15 @@ def tearDown(self):
...
-            'end_time': datetime.datetime(2019, 12, 16, 13, 45, 33), 'type': 'HDF5',
-            'collection_area_id': 'eurol'}
-
-        planned_granule_times = set([datetime.datetime(2019, 12, 13, 13, 19),
-                                     datetime.datetime(2019, 12, 13, 13, 38),
-                                     datetime.datetime(2019, 12, 13, 13, 27)])
-
-        params = {'granule_metadata': granule_metadata,
-                  'planned_granule_times': planned_granule_times}
+        params = {'granule_metadata': self.granule_metadata,
```
mraspaud, line 120: you need to remove the sensor from the metadata here, right?

TAlonglong:
Hm, I don't see this test fail when running the test locally. But I need to make some dinner now. Maybe check inn later.

pytroll_collectors/tests/test_harvest_EUM_schedules.py
```diff
@@ -117,7 +117,9 @@ def test_pass_list_file_name(self):
         self.assertEqual(os.path.join(self.basedir, 'ears_viirs_pass_prediction_19-12-16.txt'), save_file)
 
     def test_pass_list_no_sensor(self):
-        params = {'granule_metadata': self.granule_metadata,
+        granule_metadata = self.granule_metadata.copy()
+        granule_metadata.pop('sensor')
```
mraspaud, line 119: you don't need to copy the metadata, the `setUp` method is rerun before each test.

TAlonglong: Ah, ok. Let me fix that

pytroll_collectors/tests/test_region_collector.py
```diff
@@ -100,6 +125,72 @@ def test_collect_duration(europe):
     alt_europe_collector(granule_metadata)
 
 
+@unittest.mock.patch("pyorbital.tlefile.urlopen", new=_fakeopen)
+def test_collect_check_schedules(europe_collector_schedule_cut, caplog):
+    """Test default schedule cut method."""
+    granule_metadata = {
+            "platform_name": "Metop-C",
+            "sensor": "avhrr"}
```
mraspaud, line 133: this is duplicated, it could be defined outside the functions

pytroll_collectors/tests/test_region_collector.py
```diff
@@ -100,6 +125,72 @@ def test_collect_duration(europe):
...
+
+    with caplog.at_level(logging.DEBUG):
+        for s_min in (0, 3, 6, 9, 12, 15, 18):
+            europe_collector_schedule_cut.collect(
+                    {**granule_metadata,
+                     **{"start_time": datetime.datetime(2021, 4, 11, 10, s_min, 0),
+                        "end_time": datetime.datetime(2021, 4, 11, 10, s_min+3, 0),
+                        "uri": f"file://{s_min:d}"}})
```
mraspaud, line 141: these lines are duplicated, could be defined outside the functions

mraspaud (approved):
LGTM, there is just a little code duplication that is bugging me...

pytroll_collectors/tests/test_region_collector.py
TAlonglong: Hm, the start_time etc have a variable inside, s_min. I must make a function then passing the s_min return the dict.

mraspaud: ah, indeed, I missed that. But yes, a function would be nice

TAlonglong: Something like this?

mraspaud: yes, looks good!

PR #101, opened 2022-04-06
Cleanup dependencies

pnuu:
This PR cleans up dependencies of the package in the following way:
- use `pyresample.parse_area_file` instead of `satpy.get_area_def` to reduce number of additional packages
- make `pyresample`, `pytroll-schedule` and `watchdog` optional dependcies of Geographic Gatherer
- make `s3fs` optionl dependency of S3Stalker
- add `pyresample` a requirement for running unit tests where it already were required

The main motivation is to reduce the number of necessary packages installed, for example when creating a container for a single task.

The use of `pyresample.parse_area_file` means that the area definition file isn't automatically located based on `SATPY_CONFIG_PATH` environment variable, and instead `area_definition_file` needs to be defined in the configuration file of Geographic Gatherer.

In addition unused `pytroll_collectors.file_notifiers` module were removed.


pnuu:
Although the tests pass, I'll still try to run a live test with geographic gatherer to see that it really works.

pnuu:
And the latest commits should handle and report the missing config option.

pnuu:
I'm trying to test what was the effect on Docker image size for `segment_gatherer`. Seem to be getting timeouts when updating things, so might take a while.

pytroll_collectors/geographic_gatherer.py
```diff
@@ -90,7 +91,7 @@ def _setup_publisher(self):
     def _setup_triggers(self):
         """Set up the granule triggers."""
         for section in self._config.sections():
-            regions = [get_area_def(region)
+            regions = [parse_area_file(self._config.get(section, 'area_definition_file'), region)[0]
```
mraspaud, line 94: Should we default to `SATPY_CONFIG_DIR/areas.yaml` to keep some kind of backward compatibility?

mraspaud (commented):
Just one question, otherwise lgtm

pytroll_collectors/geographic_gatherer.py
pnuu: Ok, I'll add that.

pnuu:
The docker image size for `segment_gatherer` dropped from 1.97 GB to 784 MB, so quite nice improvement. The base image I'm using is already 449 MB.

pnuu:
Also the size of Docker image for `geographic_gatherer` dropped from the 1.97 GB down to 826 MB with the latest commit that makes `geographic_gatherer.py` to work solely on Posttroll messages.

mraspaud (approved):
LGTM!

PR #102, opened 2022-04-07
Make it possible to disable nameserver connections for publisher in Segment and Geographic gatherers

pnuu:
With the current version, `posttroll.publisher.NoisyPublisher` is used directly, which means it requires a `nameserver` to connect to. This PR makes it possible to disable this requirement by usage of `posttroll.publisher.create_publisher_from_dict_config()` introduced in https://github.com/pytroll/posttroll/pull/35 . When the `nameserver` connections are disabled, the message consumer connects to the publish port by defining a full address and port.

This is part of my effort to make Pytroll packages more container friendly.


pnuu:
I'll add some tests for the segment messaging stuff.

pnuu:
Forgot the `ListenerContainer`, will have a look tomorrow.

pnuu:
Converting to draft, this will need similar changes to `posttroll.listener.ListenerContainer` what was done in https://github.com/pytroll/posttroll/pull/35 for the publisher.

pnuu:
Works with https://github.com/pytroll/posttroll/pull/39 . Will require a release for the tests to pass.

pnuu:
Reminder to self: replace use of `dict_config` with `create_publisher_from_dict_config` in `pytroll_collectors/geographic_gatherer.py` and `pytroll_collectors/segments.py`.

pytroll_collectors/geographic_gatherer.py
```diff
@@ -73,17 +73,26 @@ def _clean_config(self):
...
             publisher_name = "gatherer"
 
         publish_port = self._opts.publish_port
         publisher_nameservers = self._opts.nameservers
-
-        self.publisher = publisher.NoisyPublisher(publisher_name, port=publish_port,
-                                                  nameservers=publisher_nameservers)
-        self.publisher.start()
+        if publisher_nameservers is not None:
+            publisher_nameservers = False if 'false' in publisher_nameservers else publisher_nameservers
```
mraspaud, line 89: I'd like to have just a little more error checking here: if the number of nameservers is not one, when one of them is "false", then I think an error is in order.

bin/geographic_gatherer.py
```diff
@@ -52,15 +52,17 @@ def arg_parse():
                         help="Port to publish the messages on. Default: automatic")
     parser.add_argument("-n", "--nameservers",
                         help=("Connect publisher to given nameservers: "
-                              "'-n localhost -n 123.456.789.0'. Default: localhost"),
+                              "'-n localhost -n 123.456.789.0'. Use '-n false' to disable. "
+                              "Default: localhost."
+                              ),
```
mraspaud, line 57: I'm wondering if it would be clearer to have a specific flag for this, eg `disable publisher broadcast`?

pytroll_collectors/segments.py
```diff
@@ -635,20 +636,46 @@ def _generate_publish_service_name(self):
...
+        self._publisher = create_publisher_from_dict_config(publisher_config)
+        self._publisher.start()
+
+    def _collect_publisher_config(self):
         publish_port = self._config['posttroll'].get('publish_port', 0)
         nameservers = self._config['posttroll'].get('nameservers', [])
-        services = self._config['posttroll'].get('services', "")
-        self._listener = ListenerContainer(topics=topics, addresses=addresses, services=services)
+        if nameservers is not None and 'false' in nameservers:
+            nameservers = False
```
mraspaud, line 669: this is dublicated from above (geographic_gatherer), can it be refactored?

pytroll_collectors/segments.py
```diff
@@ -635,20 +636,46 @@ def _generate_publish_service_name(self):
...
+            topics=topics,
+            addresses=addresses,
+            nameserver=nameserver,
+            services=services
+        )
+
+    def _setup_publisher(self):
+        publisher_config = self._collect_publisher_config()
+        self._publisher = create_publisher_from_dict_config(publisher_config)
+        self._publisher.start()
```
mraspaud, line 663: this is dublicated from above (geographic_gatherer), can it be refactored?

pytroll_collectors/segments.py
```diff
@@ -635,20 +636,46 @@ def _generate_publish_service_name(self):
...
         publish_service_name = self._generate_publish_service_name()
-        self._publisher = publisher.NoisyPublisher(publish_service_name,
-                                                   port=publish_port,
-                                                   nameservers=nameservers)
-        self._publisher.start()
+        publisher_config = {
+            'name': publish_service_name,
+            'nameservers': nameservers,
+            'port': publish_port,
+        }
```
mraspaud, line 677: same here.

pytroll_collectors/tests/test_geographic_gatherer.py
```diff
@@ -188,8 +211,13 @@ def test_init_posttroll(self):
...
         # A publisher is created with composed name and started
-        self.publisher.NoisyPublisher.assert_called_once_with('gatherer_'+'_'.join(sections), port=0, nameservers=None)
-        self.publisher.NoisyPublisher.return_value.start.assert_called_once()
+        expected = {
+            'name': 'gatherer_'+'_'.join(sections),
+            'port': 0,
+            'nameservers': None,
+        }
+        self.create_publisher_from_dict_config.assert_called_once_with(expected)
+        self.create_publisher_from_dict_config.return_value.start.assert_called_once()
```
mraspaud, line 220: this code is duplicated 5 times, can it be refactored into a function/method that takes `name`, `port` and `nameservers` as arguments?

pytroll_collectors/tests/test_segments.py
```diff
@@ -605,6 +605,44 @@ def test_check_and_add_existing_files(self, glob, logging):
...
+            'port': '12345',
+        }
+        creator.assert_called_with(expected_publisher)
+        creator.return_value.start.assert_called_once()
+        ListenerContainer.assert_called_once_with(
+            topics=['/foo/bar'],
+            addresses=None,
+            nameserver=False,
+            services='',
+        )
```
mraspaud, line 644: these two tests look very similar :) can they be refactored?

mraspaud (changes requested):
Thanks for adding this feature. Comments inline :)

bin/geographic_gatherer.py
pnuu: In my opinion we're not disabling broadcasting, but nameserver connection, which has the side-effect of disabling the publisher address broadcasting. I think the cleanest way would be to make "no nameserver connection" the default, and then the user would explicitly add the required nameserver if needed. But this would of course break the backwards compatibility for users who haven't always used `-n localhost` or similar. I'll think about this.

pytroll_collectors/segments.py
pnuu: Done.

pytroll_collectors/geographic_gatherer.py
pnuu: Umm, not sure how to read that. So there should be an error in this case?
```python
nameservers = ['localhost', 'false']
```

pytroll_collectors/tests/test_geographic_gatherer.py
pnuu: Done

pytroll_collectors/tests/test_segments.py
pnuu: Done

pnuu:
With the latest commits and `nameservers = [val1, val2, val3]` the following happens:
- listener is using the first one item
- if first item is `False` listener nameserver connection is disabled, otherwise nameserver connection is made
- if any of the items is `False`, publisher disables nameserver connections

mraspaud (approved):
LGTM

PR #103, opened 2022-04-22
Add return status for the run method and exit status of the main

TAlonglong:
Part of the problem in issue 100 is that the geographic gatherer exit with 0 after this error. This causes the supervisor to not restart the gatherer. 

Here I add an exit status of 1 if the exception RuntimeError or OSError is raised. This will cause the supervisor to restart the application. If this is a persisting problem, supervisor will try a configured number of times( I think 3 is the default) before it stops.

I think this at least mitigate but not solve the actual problem in 100.

### DIFF
--- a/bin/geographic_gatherer.py
+++ b/bin/geographic_gatherer.py
@@ -23,6 +23,7 @@
 
 """Gather messages of granules that cover geographic regions together and send them as a collection."""
 
+import sys
 import time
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM!

pnuu (approved):
LGTM

PR #104, opened 2022-04-22
Refactor the fsspec to message functions

mraspaud:
This PR refactors the fsspec to message functions to make them more easily reusable.

@paulovcmedeiros you might be interested in this.

This needs https://github.com/fsspec/filesystem_spec/pull/950 to work fully

### DIFF
--- a/pytroll_collectors/fsspec_to_message.py
+++ b/pytroll_collectors/fsspec_to_message.py
@@ -1,0 +1,104 @@
+"""Function to create posttroll message containing fsspec filesystem specifications."""
+
+import json
+import os
+
+from fsspec import get_filesystem_class
+from posttroll.message import Message
+
+
+def create_message(fs, file, subject, metadata=None):
+    """Create a message to send."""
+    if isinstance(file, (list, tuple)):
+        file_data = {'dataset': []}
+        for file_item in file:
+            file_data['dataset'].append(_create_message_metadata(fs, file_item))
+        message_type = 'dataset'
+    else:
+        file_data = _create_message_metadata(fs, file)
+        message_type = 'file'
+    if metadata:
+        file_data.update(metadata)
+    return Message(subject, message_type, file_data)
+
+
+def _create_message_metadata(fs, file):
+    """Create a message to send."""
+    loaded_fs = json.loads(fs)
+    uri = _create_uri(file, loaded_fs)
+    uid = _create_uid_from_uri(uri, loaded_fs)
+    base_data = {'filesystem': loaded_fs, 'uri': uri, 'uid': uid}
+    base_data.update(file.get('metadata', dict()))
+    return base_data
+
+
+def _create_uri(file, loaded_fs):
+    protocol = loaded_fs["protocol"]
+    if protocol == 'abstract' and 'zip' in loaded_fs['cls']:
+        protocol = 'zip'
+    uri = protocol + '://' + file['name']
+    return uri
+
+
+def _create_uid_from_uri(uri, loaded_fs):
+    uid = uri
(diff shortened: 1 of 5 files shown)


pnuu (approved):
LGTM, feel free to merge when the fsspec PR is merged and released so this can be actually used.

pnuu:
Please remind me, what was the use for the `filesystem` part of the message dictionary? As far as I can see it isn't needed by `fsspec`:
```python
uri = "tar://tmp/bar.txt::ssh://127.0.0.1/tmp/test.tar"
fsfiles = fsspec.open_files(uri)
print(fsfiles[0])
# <OpenFile 'tmp/bar.txt'>
```
And what does `fo` stand for? If "output file", then isn't the tar file actually the input where the data is coming from and then stored locally as `uid`?

mraspaud:
the `filesystem` item is just here to make sure we have everything. Indeed, if we were to include eg `filecache`, the specific options for it can't be added to the uri, but need to be provided as storage options inside `filesystem`. Now, it is true that the implemented functions do not add filecache yet...

Regarding `fo`, from what I gather it means `file object`.

mraspaud:
@pnuu there was a significant change in the last commit, maybe you want to have a look at it again.

pytroll_collectors/fsspec_to_message.py
```diff
@@ -25,31 +26,50 @@ def create_message(fs, file, subject, metadata=None):
...
+    return uri
+
+
+def create_netloc(ssh):
+    """Create the netloc from ssh parameters like `host`, `port`, `username`, `password`."""
+    host = ssh.get("host")
+    if host is None:
+        return ""
+    username = ssh.get("username")
+    password = ssh.get("password")
```
pnuu, line 59: I hope this doesn't imply that the credentials are passed in plain text in the messages everyone within the network can see...?

pytroll_collectors/fsspec_to_message.py
```diff
@@ -25,31 +26,50 @@ def create_message(fs, file, subject, metadata=None):
...
-    uri = protocol + '://' + file['name']
-    return uri
+    netloc = create_netloc(loaded_fs)
+    uid = protocol + '://' + netloc + file['name']
+    return uid
 
 
-def _create_uid_from_uri(uri, loaded_fs):
-    uid = uri
+def _create_uid_from_uri(uid, loaded_fs):
```
pnuu, line 45: This is creating URI from UID, so the naming is reversed

pytroll_collectors/fsspec_to_message.py
```diff
@@ -84,7 +104,34 @@ def extract_files_to_message(file, fs, subject, packing=None):
...
+    if packing is None:
+        cls = get_filesystem_class(target_protocol or "ssh")
+        cls = ".".join((cls.__module__, cls.__name__))
+        target_options = target_options or dict(host=socket.gethostname())
+        fs_dict = dict(cls=cls,
+                       protocol=target_protocol or "ssh",
+                       args=[],
+                       **target_options
+                       )
+        file = dict(name=filename)
```
pnuu, line 125: This is not tested

pnuu (changes requested):
One misnamed function, one security concern and some lines the tests don't cover.

pytroll_collectors/fsspec_to_message.py
mraspaud: Fixed

pytroll_collectors/fsspec_to_message.py
mraspaud: Fixed

pnuu (approved):
LGTM

PR #107, opened 2022-05-20
Fix too many protocols

mraspaud:
The recent changes for allowing the creation of uris for remote filesystems was not sufficiently tested, and real world tests exposed a bug.

This fixes it.


pnuu (approved):
LGTM

PR #109, opened 2022-07-01
Confirm configuration file was read in gatherer.

### LINKED ISSUE #72: gatherer does not complain when configuration file does not exist
Summary
---

When the configuration file passed to gatherer does not exist (or otherwise cannot be read), there is neither a warning nor an error from gatherer.py.  From the console output (or logfile output) there is no direct indication that anything is wrong — only the indirect evidence from the observation that it doesn't seem to be doing much.

MCVE
----

```
$ gatherer.py -v /does/not/exist
```

Expected output
----

I think this should fail immediately, as I don't see any way in which this error could be recoverable.

Actual output
----

```
$ gatherer.py -v /does/not/exist
Setting timezone to UTC
[INFO: 2021-01-14 15:02:24 : posttroll.publisher] publisher started on port 38155
```

gerritholl:
Confirm the configuration file was read successfully when initialising
the geographic gatherer.

Fixes #72.

### DIFF
--- a/pytroll_collectors/geographic_gatherer.py
+++ b/pytroll_collectors/geographic_gatherer.py
@@ -45,7 +45,11 @@ class GeographicGatherer:
     def __init__(self, opts):
         """Initialize the class."""
         self._config = ConfigParser(interpolation=None)
-        self._config.read(opts.config)
+        # ConfigParser.read silently ignores unreadable files...
+        read = self._config.read(opts.config)
+        if not read:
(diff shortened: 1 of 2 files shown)


gerritholl:
I'm not authorised to request reviews here, but maybe @mraspaud @TAlonglong @pnuu can try this out in their workflow?  It does have the potential to break compatibility in theory, but I don't think it should in practice.

pnuu (approved):
LGTM

PR #111, opened 2022-07-13
When adding to existing timeslot, indicate which one

gerritholl:
In the debug message informing that an existing timeslot had been found,
indicate what timeslot this was.


mraspaud (commented):
LGTM

PR #114, opened 2022-08-17
Add an s3stalker daemon

adybbroe:
Add a runner or demon to stalk for new files in an S3 bucket based on the s3stalker script


bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+            wait_seconds = waiting_time.total_seconds()
+
+            # Wait for some time...
+            logger.debug("Waiting %d seconds", wait_seconds)
+            time.sleep(max(wait_seconds, 0))
+
+    def _get_seconds_back_to_search(self, last_fetch_time):
+        """Update the time to look back considering also the modification time of the last file."""
+        dtime_back = timedelta(**self.time_back)
+        seconds_back = dtime_back.total_seconds()
```
pnuu, line 109:
```suggestion
        seconds_back = timedelta(**self.time_back).total_seconds()
```

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+        while self.loop:
+            if first_run:
+                logger.info('Create messages with urls for most recent files only')
+                logger.info('On start up we consider files with age up to %d seconds from now',
+                            self.startup_timedelta_seconds)
+                config['timedelta'] = {'seconds': self.startup_timedelta_seconds}
+                first_run = False
+            else:
+                seconds_back = self._get_seconds_back_to_search(last_fetch_time)
+                config['timedelta'] = {'seconds': seconds_back}
```
pnuu, line 88: This could be refactored to a separate method, and used something like this:
```python
config['timedelta'] = self._get_timedelta(last_fetch_time, is_first_run=first_run)
first_run = False
```

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def close(self):
+        """Shutdown the S3Stalker runner."""
+        logger.info('Terminating the S3 Stalker daemon/runner.')
+        self.loop = False
+        if self.publisher:
+            try:
+                self.publisher.stop()
+            except Exception:
+                logger.exception("Couldn't stop publisher.")
```
pnuu, line 127: I think this try/except isn't needed, the `if self.publisher` check handles the case where it is `None`, and otherwise it is a safe operation. At least there are no try/except in any of the places it is used I checked :-)

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+            log_dict = yaml.safe_load(fd.read())
+            logging.config.dictConfig(log_dict)
+
+    logger.info("Try start the s3-stalker runner:")
+    try:
+        s3runner = S3StalkerRunner(bucket, config, startup_timedelta_seconds)
+
+    except Exception as err:
+        logger.error('The S3 Stalker Runner crashed: %s', str(err))
+        sys.exit(1)
```
pnuu, line 165: I think is a bit excessive when checking that only the initialization is failing and the stalker hasn't even yet started. Maybe merge this and the next try/except block?

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+        """Initialize the S3Stalker runner class."""
+        super().__init__()
+
+        self.bucket = bucket
+        self.config = config
+        self.startup_timedelta_seconds = startup_timedelta_seconds
+        self.time_back = self.config['timedelta']
+        self.publisher = None
+        self.loop = False
+        self._setup_and_start_communication()
```
pnuu, line 60: Would it make sense to automatically start the stalker here? Or move the publisher creation in `.run()`? Now `.close()` call is needed even if the stalker hasn't been started, and that might not be so evident.

pnuu (changes requested):
Some comments, questions and suggestions :-)

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,176 @@
...
+from pytroll_collectors.s3stalker import get_last_fetch
+
+import signal
+from threading import Thread
+from posttroll.publisher import create_publisher_from_dict_config
+
+logger = logging.getLogger(__name__)
+
+
+class S3StalkerRunner(Thread):
```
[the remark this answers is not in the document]
adybbroe: Yes, right. Move it

bin/s3stalker_daemon.py
adybbroe: Yes, you are right! Didn't see that. I have moved the start of the publisher in under the `run` method now.

bin/s3stalker_daemon.py
adybbroe: Done

bin/s3stalker_daemon.py
adybbroe: Done

bin/s3stalker_daemon.py
adybbroe: Ok, I see. You are probably right.

bin/s3stalker_daemon.py
adybbroe: Right, thanks, done

adybbroe:
Damn. Moving the `self._setup_and_start_communication()`from the `init` to the `run` method as suggested was a bad idea I am afraid, it gives me this error:
`ValueError: signal only works in main thread of the main interpreter`
So, I am moving it back into the init @pnuu

adybbroe:
> Damn. Moving the `self._setup_and_start_communication()`from the `init` to the `run` method as suggested was a bad idea I am afraid, it gives me this error: `ValueError: signal only works in main thread of the main interpreter` So, I am moving it back into the init @pnuu

Oh, I see now, that you only suggested moving the publisher creation to the run method, sorry will redo...

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+            logger.info("Publishing %s", str(message))
+            self.publisher.send(str(message))
+
+    def _get_timedelta(self, last_fetch_time, is_first_run):
+        """Get the seconds for the time window to search for (new) files."""
+        if is_first_run:
+            logger.info('Create messages with urls for most recent files only')
+            logger.info('On start up we consider files with age up to %d seconds from now',
+                        self.startup_timedelta_seconds)
+            return {'seconds': self.startup_timedelta_seconds}
```
pnuu, line 108: The `datetime.timedelta()` function is only given seconds, so to simplify just return the value without a dictionary, and use `timedelta(seconds=return_value_from_here)`.

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+
+        seconds_back = self._get_seconds_back_to_search(last_fetch_time)
+        return {'seconds': seconds_back}
+
+    def _get_seconds_back_to_search(self, last_fetch_time):
+        """Update the time to look back considering also the modification time of the last file."""
+        seconds_back = timedelta(**self.time_back).total_seconds()
```
pnuu, line 115: Oh. Isn't this the same as just using the seconds directly without using `timedelta()` at all? So no need for returning dictionaries and nor going from seconds to `timedelta` to seconds. Right?

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+
+        if last_fetch_time is None:
+            return seconds_back
+
+        start_time = datetime.utcnow()
+        start_time = start_time.replace(tzinfo=timezone.utc)
```
pnuu, line 121: The time is already in UTC...?

pytroll_collectors/s3stalker.py
```diff
@@ -54,6 +149,7 @@ class DatetimeHolder:
 
 def get_last_files(path, *args, pattern=None, **kwargs):
     """Get the last files from path (s3 bucket and directory)."""
+    kwargs['skip_instance_cache'] = True
```
pnuu, line 152: Um, what? Why is a kwarg modified here? Should the value be an instance attribute of the class? Then there could be a wrapper method that adjusts the attribute and calls this function.

pytroll_collectors/s3stalker.py
```diff
@@ -93,17 +189,33 @@ def set_last_fetch(timestamp):
...
+
+
+def create_messages_for_recent_files(bucket, config):
+    """Create messages for recent files and return."""
+    logger.debug("Create messages for recent files...")
+    time_back = config['timedelta']
+    logger.debug("time_back = %s", str(time_back))
+    subject = config['subject']
+    pattern = config.get('file_pattern')
+    with sleeper(2.5):
```
pnuu, line 204: Is it important that the code block should take at least 2.5 seconds to process?

pytroll_collectors/s3stalker.py
```diff
@@ -93,17 +189,33 @@ def set_last_fetch(timestamp):
...
+        set_last_fetch(datetime.now(tz.UTC) - timedelta(**time_back))
```
pnuu, line 205:
```suggestion
        set_last_fetch(datetime.utcnow()) - timedelta(**time_back))
```

pytroll_collectors/s3stalker.py
```diff
@@ -93,17 +189,33 @@ def set_last_fetch(timestamp):
...
+        s3_kwargs = config['s3_kwargs']
+        logger.debug('s3_kwargs: %s', str(s3_kwargs))
```
pnuu, line 207: This will print also the credentials, right? Not good if so...

pnuu (changes requested):
This seems to be (at least) three pull requests in one:
- CI changes
- S3stalker changes
- scisys receiver changes

Are the two here on purpose?

I added some code comments. I think I'll need to look at the whole code to give more structural comments, but that needs to be after the holidays.

.github/workflows/ci.yaml
```diff
@@ -1,36 +1,86 @@
...
-    runs-on: ubuntu-latest
+  test:
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        os: ["ubuntu-latest"]
+        python-version: ["3.9", "3.10"]
```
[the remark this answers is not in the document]
adybbroe: Yes, you are right, fixed now.

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+        self.loop = False
+        self._set_signal_shutdown()
+
+    def _set_signal_shutdown(self):
+        """Set a signal to handle shutdown."""
+        signal.signal(signal.SIGTERM, self.signal_shutdown)
+
+    def _setup_and_start_communication(self):
+        """Set up the Posttroll communication and start the publisher."""
+        logger.debug("Starting up... ")
```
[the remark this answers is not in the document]
adybbroe: Yes, good suggestion, fixed, thx.

pytroll_collectors/s3stalker.py
adybbroe: Hmm, you are right

pytroll_collectors/s3stalker.py
adybbroe: Yes, think you are right, will remove

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+
+    def run(self):
+        """Start the s3-stalker daemon/runner in a thread."""
+        self._setup_and_start_communication()
+
+        waiting_time = timedelta(**self.time_back)
+        wait_seconds = waiting_time.total_seconds()
+
+        first_run = True
+        config = self.config
```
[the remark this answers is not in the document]
adybbroe: Done

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+
+    def do_fetch_most_recent(self, last_fetch_time, first_run, config):
+        """Do a fecth of the most recent files and create and send the messages."""
+        config['timedelta'] = self._get_timedelta(last_fetch_time, is_first_run=first_run)
+
+        last_fetch_time = get_last_fetch()
+        logger.info("Last fetch time...: %s", str(last_fetch_time))
+
+        self._process_messages(config)
+        return last_fetch_time
```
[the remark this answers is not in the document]
adybbroe: Yes, it was easier testing this way I thought, but actually not really. I have done as suggested.

pytroll_collectors/s3stalker.py
```diff
@@ -23,37 +19,133 @@
...
+    """Runner for stalking for new files in an S3 object store."""
+
+    def __init__(self, bucket, config, startup_timedelta_seconds):
+        """Initialize the S3Stalker runner class."""
+        super().__init__()
+
+        self.bucket = bucket
+        self.config = config
+        self.startup_timedelta_seconds = startup_timedelta_seconds
+        self.time_back = self.config['timedelta']
```
pnuu, line 49: This is used only in forming `self._wait_seconds` below, so not needed as an attribute.

pytroll_collectors/s3stalker.py
```diff
@@ -23,37 +19,133 @@
...
+class S3StalkerRunner(Thread):
+    """Runner for stalking for new files in an S3 object store."""
+
+    def __init__(self, bucket, config, startup_timedelta_seconds):
+        """Initialize the S3Stalker runner class."""
+        super().__init__()
+
+        self.bucket = bucket
+        self.config = config
+        self.startup_timedelta_seconds = startup_timedelta_seconds
```
pnuu, line 48: I find the code a bit hard to follow, but as user-configured `self._timedelta` is adjusted (why?), could this and `self._timedelta` attributes merged into only one attribute?

pytroll_collectors/s3stalker.py
```diff
@@ -23,37 +19,133 @@
...
+        self.publisher.start()
+        self.loop = True
+
+    def signal_shutdown(self, *args, **kwargs):
+        """Shutdown the S3 Stalker daemon/runner."""
+        self.close()
+
+    def run(self):
+        """Start the s3-stalker daemon/runner in a thread."""
+        logger.info("Starting up... ")
```
pnuu, line 73:
```suggestion
        logger.info("Starting up s3stalker.")
```

pytroll_collectors/s3stalker.py
```diff
@@ -23,37 +19,133 @@
...
+        self._setup_and_start_communication()
+
+        first_run = True
+        last_fetch_time = None
+        while self.loop:
+            self._set_timedelta(last_fetch_time, first_run)
+
+            last_fetch_time = get_last_fetch()
+            logger.info("Last fetch time...: %s", str(last_fetch_time))
```
pnuu, line 82:
```suggestion
            logger.debug("Last fetch time...: %s", str(last_fetch_time))
```

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+        """Go through all messages in list and publish them one after the other."""
+        messages = create_messages_for_recent_files(self.bucket, config)
+        for message in messages:
+            logger.info("Publishing %s", str(message))
+            self.publisher.send(str(message))
+
+    def _get_timedelta(self, last_fetch_time, is_first_run):
+        """Get the seconds for the time window to search for (new) files."""
+        if is_first_run:
+            logger.info('Create messages with urls for most recent files only')
```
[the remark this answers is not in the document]
pnuu: And in any case this should be a debug message.

pytroll_collectors/s3stalker.py
pnuu: The times seem to be always UTC, so there's no specific need for being timezone aware.

pytroll_collectors/s3stalker.py
pnuu: Any comments here?

pytroll_collectors/s3stalker.py
pnuu: Comments/reasoning for this?

pytroll_collectors/s3stalker.py
```diff
@@ -93,18 +185,31 @@ def set_last_fetch(timestamp):
...
+        messages = filelist_unzip_to_messages(fs_, files, subject)
+        # fs_.clear_instance_cache()
+
+    return messages
+
+
+def publish_new_files(bucket, config, time_back):
     """Publish files newly arrived in bucket."""
+    messages = create_messages_for_recent_files(bucket, config, time_back)
     with Publish("s3_stalker") as pub:
```
pnuu, line 212: The publisher should be configurable for example not to use nameserver, allow the user to set the publisher port, etc.

pytroll_collectors/s3stalker.py
```diff
@@ -23,37 +19,133 @@
...
     waiting_time = duration - (end_time - start_time).total_seconds()
+    logger.debug('waiting time: %f', waiting_time)
     time.sleep(max(waiting_time, 0))
 
 
 class DatetimeHolder:
     """Holder for the last_fetch datetime."""
 
-    last_fetch = datetime.now(tz.UTC) - timedelta(hours=12)
+    last_fetch = datetime.utcnow() - timedelta(hours=12)
```
pnuu, line 143: If I read correctly, this class is used as a globla variable, right? Could the `last_fetch` time be attribute of the `S3StalkerRunner` class? That way there would be no need for `get/set_last_fetch()` functions and `get_last_files()` could be a method without `*args` and `**kwargs`.

pnuu (changes requested):
Comments inline.

pytroll_collectors/s3stalker.py
```diff
@@ -23,26 +19,125 @@
...
+    def _set_signal_shutdown(self):
+        """Set a signal to handle shutdown."""
+        signal.signal(signal.SIGTERM, self.signal_shutdown)
+
+    def _setup_and_start_communication(self):
+        """Set up the Posttroll communication and start the publisher."""
+        logger.debug("Starting up... ")
+        self.publisher = create_publisher_from_dict_config(self.config['publisher'])
+        self.publisher.start()
+        self.loop = True
```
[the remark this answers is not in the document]
adybbroe: @pnuu I didn't figure out how to properly test code in an infinite loop, please help!

pytroll_collectors/s3stalker.py
adybbroe: I am using `freezegun.freeze_time` for the tests, which I think was the reason behind the zone awareness. Can we perhaps keep the time zone awareness, for now? Or can you make suggestions on how to test it without?

pytroll_collectors/s3stalker.py
adybbroe: Do you remember why we needed this @TAlonglong ? 
The `skip_instance_cache` was apparently needed in order to get this to work when we polled for data in the bucket of @TAlonglong 
I think it is fair not to require it as user-configurable item, so don't want it in the config. I am unsure what you are after @pnuu ?

pytroll_collectors/s3stalker.py
adybbroe: Not sure about the exact amount of `2.5` how that was achieved. Do you remember @TAlonglong ?
I can move it into a configuration item, would that be okay?

pytroll_collectors/s3stalker.py
adybbroe: Help wanted! @mraspaud Did this code

pytroll_collectors/s3stalker.py
adybbroe: Actually git-blame says it was added by @mraspaud

pytroll_collectors/s3stalker.py
adybbroe: Could we do that in another PR (or issue) maybe? It was nothing I touched as far as I know/remember?

pnuu: Certainly! Sorry, switching between different things got confused and thought all of the s3stalker stuff was implemented in this PR :man_facepalming:

pytroll_collectors/s3stalker.py
pnuu: We can move that to another PR, as I commented just before I was somehow confused what already existed and what didn't.

pytroll_collectors/s3stalker.py
```diff
@@ -93,18 +185,31 @@ def set_last_fetch(timestamp):
     DatetimeHolder.last_fetch = timestamp
 
 
-def publish_new_files(bucket, config):
+def get_last_fetch():
+    """Get the last fetch time."""
+    return DatetimeHolder.last_fetch
+
+
+def create_messages_for_recent_files(bucket, config, time_back):
```
mraspaud, line 193: Why does `time_back` need to be passed here? it is available in the config, right?

mraspaud (changes requested):
Some comments, but I'm going to have a closer look, probably in a PR to this PR.

pytroll_collectors/s3stalker.py
mraspaud: the problem with `utcnow` is that it doesn't contain timezone info, so once you get it, you can't track what timezone it is in. Since the file list we get through `s3fs` returns file timestamps with a timezone information, we need to have it in our datetime objects too to be comparable.

```python
In [8]: datetime.now(tz.UTC)
Out[8]: datetime.datetime(2023, 2, 14, 12, 54, 19, 214202, tzinfo=tzutc())

In [9]: datetime.utcnow()
Out[9]: datetime.datetime(2023, 2, 14, 12, 54, 24, 733345)
```

pytroll_collectors/s3stalker.py
```diff
@@ -69,7 +91,8 @@ def _reset_last_fetch_from_file_list(files):
 
 def _get_files_since_last_fetch(fs, path):
     files = fs.ls(path, detail=True)
-    files = list(filter((lambda x: x['LastModified'] > DatetimeHolder.last_fetch), files))
+    logger.debug(f"Get files since {get_last_fetch}")
```
pnuu, line 94: I think this should be
```suggestion
    logger.debug(f"Get files since {get_last_fetch()}")
```
The original would print `<function get_last_fetch at 0x....>`:

```python
In [1]: def foo(): return 'asd'

In [2]: print(f"{foo}")
<function foo at 0x7ff63872a7a0>

In [3]: print(f"{foo()}")
asd
```

pytroll_collectors/s3stalker.py
```diff
@@ -88,23 +111,26 @@ def _match_files_to_pattern(files, path, pattern):
...
         for message in messages:
             logger.info("Publishing %s", str(message))
             pub.send(str(message))
+
+
+def create_messages_for_recent_files(bucket, config):
+    """Create messages for recent files and return."""
+    logger.debug("Create messages for recent files...")
+
+    logger.debug(f"Last fetch is {get_last_fetch()}")
```
pnuu, line 130: This message is logged also when the file list is retrieved, one of them should be removed.

pytroll_collectors/s3stalker.py
```diff
@@ -88,23 +111,26 @@ def _match_files_to_pattern(files, path, pattern):
...
+    subject = config['subject']
+    pattern = config.get('file_pattern')
+    s3_kwargs = config['s3_kwargs']
+    fs_, files = get_last_files(bucket, pattern=pattern, **s3_kwargs)
```
pnuu, line 134: I think it would be cleaner if this was outside of `create_messages_for_recent_files()`. And the `get_last_files()` could even be given just the config. The remaining call to `filelist_unzip_to_messages()` could then be renamed to `create_messages_for_recent_files()` to replace this function.

pytroll_collectors/s3stalker_daemon_runner.py
```diff
@@ -0,0 +1,104 @@
...
+
+class S3StalkerRunner(Thread):
+    """Runner for stalking for new files in an S3 object store."""
+
+    def __init__(self, bucket, config, publisher_ready_time=2.5):
+        """Initialize the S3Stalker runner class."""
+        super().__init__()
+
+        self.bucket = bucket
+        startup_time = timedelta(**config.pop("fetch_back_to"))
```
pnuu, line 52: This isn't startup time, so the variable name is misleading.

The same config item is accessed in `s3stalker.py` function `publish_new_files()`. Is the daemon passing the same config dictionary to it or are these independent? If the same config is passed, the other is passing `None` to `timedelta()`.

pytroll_collectors/s3stalker_daemon_runner.py
```diff
@@ -0,0 +1,104 @@
...
+        self._start_communication()
+
+        while self.loop:
+            self._fetch_bucket_content_and_publish_new_files()
+
+            self._wait_until_next_poll()
+
+    def _wait_until_next_poll(self):
+        """Wait until it's time for next poll."""
+        wait_time = max(self._wait_seconds, 0)
```
pnuu, line 88: As I see it, the `max()` is not needed, `self._wait_seconds` is coming from the config and as-implemented will either be a valid number of raise `TypeError` if `polling_interval` isn't present in the config. So `self._wait_seconds` can be used directly in the log message and sleep call.

setup.py
```diff
@@ -89,8 +89,9 @@
       install_requires=['posttroll>=1.3.0',
                         'trollsift',
                         'pyyaml'],
-      tests_require=['trollsift', 'netifaces', 'watchdog', 'posttroll', 'pyyaml', 'pyinotify', 's3fs',
+      tests_require=['trollsift', 'netifaces', 'watchdog', 'posttroll', 'pyyaml',
+                     'pyinotify', 's3fs', 'freezegun',
                      'pyresample', 'python-dateutil', 'posttroll', 'pytest'],
       extras_require=extras_require,
-      python_requires='>=3.7',
+      python_requires='>=3.8',
```
pnuu, line 96: The changes for CI is dropping 3.8, so should this also have 3.9?

pnuu (commented):
Some new comments.

doc/source/index.rst
```diff
@@ -456,6 +456,21 @@ trollstalker2
 New, alternative implementation of trollstalker.  Not really needed,
 as trollstalker works fine and is actively maintained.
 
+
+s3stalker
+^^^^^^^^^
+
+A counterpart to trollstalker for polling for new files on an s3 bucket.
+This is though to be run regularly from eg. cron. For a daemon version of
```
mraspaud, line 464:
```suggestion
This is thought to be run regularly from eg. cron. For a daemon version of
```

pytroll_collectors/s3stalker_daemon_runner.py
mraspaud: Indeed, this is something I forgot to clean up from the previous version.
Once again, refactoring made the error pop out better :)

pytroll_collectors/s3stalker.py
mraspaud: I agree that this is strange. I can see that the kwargs are coming straight from the config file's `s3_kwargs`, so this should be a configuration item, not hardcoded here.

setup.py
mraspaud: yes

pytroll_collectors/s3stalker_daemon_runner.py
mraspaud: Fixed

pytroll_collectors/s3stalker_daemon_runner.py
mraspaud: Fixed the name

mraspaud: `publish_new_files` is not called by the daemon, since this is made to be called once and exit.

pytroll_collectors/s3stalker.py
mraspaud: done

pytroll_collectors/s3stalker.py
mraspaud: If that's ok with you, I'd like to keep this as it is. I feel passing the config to `get_last_file` would make it's call less clear, while now we can read clearly that this function gets the last file from a bucket with the pattern and these s3 parameters.
I also this this would clutter the tests more than they already are...

bin/s3stalker_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def arg_parse():
+    """Handle input arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("bucket", help="The bucket to retrieve from.")
+    parser.add_argument("config", help="Config file to be used")
+    parser.add_argument("-l", "--log",
+                        help="Log configuration file",
+                        default=None)
```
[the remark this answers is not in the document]
mraspaud: I don't think the log config belongs in the main configuration file, I believe it's best to keep them separated.
Regarding the bucket, I think it's a natural way to call the command, makes it clear what we are stalking. But if you feel strongly about it, we can change that.

mraspaud:
@pnuu I have addressed or answered your comments...

bin/s3stalker_daemon.py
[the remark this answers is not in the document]
pnuu: I'm ok to leave the log config as separate file. Personally I like to have all the configurations in the config file, that's way they are self contained and easier to maintain. So how about allow it to be there, and let the command-line argument to override it if given?

pytroll_collectors/s3stalker.py
pnuu: Yeah, I was thinking about that too. Having distinct arguments makes sense.

But what about moving `get_last_files()` out of the `create_new_messages_for_recent_files()` and renaming `filelist_unzip_to_messages()`?

mraspaud: `filelist_unzip_to_messages` hasn't been touched in this PR and has a logical name for what it's doing IMO, that is converting a list of files (possibly zipped) to posttroll messages (see here https://github.com/pytroll/pytroll-collectors/blob/main/pytroll_collectors/fsspec_to_message.py#L77-L89). `create_new_messages_for_recent_files` on the other hand assumes that the files are recent, but that can only be the case if we call `get_last_files` first...

pnuu:
I went through the unresolved comments, and only my two latest ones are still open.

For some reason GitHub shows some of the comments without the ability to resolve them, so the comment section is a bit messy.

doc/source/index.rst
```diff
@@ -456,6 +456,21 @@ trollstalker2
...
+
+A counterpart to trollstalker for polling for new files on an s3 bucket.
+This is thought to be run regularly from eg. cron. For a daemon version of
+this, check the next item.
+
+s3stalker_daemon
+^^^^^^^^^^^^^^^^
+
+The daemon version of s3stalker, that stays on and polls until stopped
+(preferably with a SIGTERM).
```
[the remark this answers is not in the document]
mraspaud: Ok, I added a link to the example config. I'm not familiar with the boto arguments, so if you want to have something, maybe add a suggestion here that we can merge?

doc/source/index.rst
```diff
@@ -463,13 +463,16 @@ s3stalker
...
+https://github.com/pytroll/pytroll-collectors/blob/main/examples/s3stalker.yaml
 
 s3stalker_daemon
 ^^^^^^^^^^^^^^^^
 
 The daemon version of s3stalker, that stays on and polls until stopped
 (preferably with a SIGTERM).
-
+Example configuration:
+https://github.com/pytroll/pytroll-collectors/blob/main/examples/s3stalker_runner.yaml_template
```
pnuu, line 475: Maybe something as simple as this?

```suggestion
https://github.com/pytroll/pytroll-collectors/blob/main/examples/s3stalker_runner.yaml_template

See also https://s3fs.readthedocs.io/en/latest/#credentials on options how to define the S3 credentials.
```

pnuu (approved):
LGTM, thanks for the updates!

Just one suggestion on the S3 credential configuration.

bin/s3stalker_daemon.py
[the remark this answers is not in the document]
pnuu: Bucket-in-config was added in https://github.com/pytroll/pytroll-collectors/pull/114/commits/e9b6f3eba4b86d077cddcc1ddc060a2dab9378f7

Issue #30, opened 2019-02-14
If tle is missing the actual gatherer thread is crashing


TAlonglong:
If tle for some reason is missing the actual gatherer thread is crashing like this:

```[DEBUG: 2019-02-13 04:12:02 : pyorbital.tlefile] Reading TLE from /data/pytroll/tle-in/tle_db/tle-latest-pytroll.txt
[DEBUG: 2019-02-13 04:12:02 : trollsched.satpass] Failed in PyOrbital: u"Found no TLE entry for 'NOAA 19'"
[DEBUG: 2019-02-13 04:12:02 : pyorbital.tlefile] Reading TLE from /data/pytroll/tle-in/tle_db/tle-latest-pytroll.txt
Exception in thread Thread-11:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/trigger.py", line 381, in run
    self.process(msg)
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/trigger.py", line 94, in add_file
    self._do(pathname)
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/trigger.py", line 89, in _do
    Trigger._do(self, mda)
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/trigger.py", line 64, in _do
    res = collector(metadata.copy())
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/region_collector.py", line 61, in __call__
    return self.collect(granule_metadata)
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_collectors-0.7.0-py2.7.egg/pytroll_collectors/region_collector.py", line 169, in collect
    instrument=granule_metadata["sensor"])
  File "/software/pytroll/lib/python2.7/site-packages/pytroll_schedule-0.3.3-py2.7.egg/trollsched/satpass.py", line 184, in __init__
    self.orb = orbital.Orbital(NOAA20_NAME.get(satellite, satellite), line1=tle1, line2=tle2)
[...]

PR #116, opened 2022-09-01
Catch KeyError from collector from missing TLE

TAlonglong:
Now and then I get for some reason an exception as described in #30

What happens is then that the subscriber for this area stops and thus can not continue collecting new segments.
Other regions continue to work if there are more. Or the the geographic gatherer just continue to run.

One way to handle this, as proposed in this PR, is to catch the KeyError exception and then continue keeping the subscriber running. Therefore at the next run it will try again and it will still work.

I think this is better than just stop working. At least it tries again.

Another way is to do this is to make the whole geographic gatherer stop and then leave to the user how he or she will handle this. But this will affect all the regions in the gatherer.

### DIFF
--- a/pytroll_collectors/tests/test_triggers.py
+++ b/pytroll_collectors/tests/test_triggers.py
@@ -264,3 +264,29 @@ class TestFileTrigger:
                        'uri': 'somefile_20220512T1544_20220512T1545.data',
                        'key1': "value1"
                        }
+
+
+    @patch('pytroll_collectors.triggers._base.FileTrigger')
+    def test_filetrigger_exception(self, patch_filetrigger):
+        """Test getting the metadata."""
+        from pytroll_collectors.triggers._base import FileTrigger
+        import configparser
+        def _collectors(metadata):
+            raise KeyError("Found no TLE entry for 'METOP-B'")
+        collectors = [_collectors]
+        config = configparser.ConfigParser(interpolation=None)
+        section = "section1"
+        config.add_section(section)
+        config.set(section, "pattern", "{name}_{start_time:%Y%m%dT%H%M}_{end_time:%Y%m%dT%H%M}.data")
+        publisher = None
+        trigger = FileTrigger(collectors, dict(config.items("section1")), publisher, publish_topic=None,
+                              publish_message_after_each_reception=False)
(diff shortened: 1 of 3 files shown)


pytroll_collectors/tests/test_triggers.py
```diff
@@ -264,3 +264,48 @@ def test_getting_metadata_does_not_involve_multiple_sections(self):
...
+        trigger = FileTrigger(collectors, dict(config.items("section1")), publisher, publish_topic=None,
+                              publish_message_after_each_reception=False)
+        trigger._process_metadata({'sensor': 'avhrr', 'platform_name': 'Metop-B',
+                                   'start_time': datetime(2022, 9, 1, 10, 22, 3),
+                                   'orbit_number': '51653',
+                                   'uri': 'AVHRR_C_EUMP_20220901102203_51653_eps_o_amv_l2d.bin',
+                                   'uid': 'AVHRR_C_EUMP_20220901102203_51653_eps_o_amv_l2d.bin',
+                                   'origin': '157.249.16.188:9062',
+                                   'end_time': datetime(2022, 9, 1, 10, 25, 3)})
+        assert "Found no TLE entry for 'METOP-B' to similate KeyError" in caplog.text
```
pnuu, line 290:
```suggestion
        assert "Found no TLE entry for 'METOP-B' to simulate" in caplog.text
```

pnuu: Hmm, where does this exception text originate? I found `"Found no TLE entry for '%s'"` in https://github.com/pytroll/pyorbital/blob/main/pyorbital/tlefile.py#L200 but the rest of it I can't seem to find in either Pyorbital, Pytroll Schedule nor Pytroll Collectors.

pnuu (commented):
In general LTGM, just a question on the test of raised `KeyError`.

pytroll_collectors/tests/test_triggers.py
TAlonglong: oh, how to explain

I just added some dummy text in the test to be sure I matched the test KeyError message and not the message from the actual code. If that make any sense. I struggled a bit before I got this right.

But sure, I can at least get the spelling right.

pnuu: Oh, right. Didn't grep the same file :sweat_smile:

PR #110, opened 2022-07-13
Improving segment gatherer documentation

gerritholl:
Improve the documentation for the segment gatherer, in particular for the configuration.

PR #117, opened 2022-09-05
Check for existing segments on S3 storage

pnuu:
This PR adds a possibility to check for existing file segments on S3 storage when segment gatherer is started. This is already possible with local file systems by defining `check_existing_files_after_start = True` in the configuration file.

Not sure where to explain how the S3 configuration (end point, credential) is done.

### DIFF
--- a/pytroll_collectors/tests/test_segments.py
+++ b/pytroll_collectors/tests/test_segments.py
@@ -622,52 +622,58 @@ class TestSegmentGatherer(unittest.TestCase):
 
     @patch("pytroll_collectors.segments.logging")
     @patch("fsspec.filesystem")
-    def test_check_and_add_existing_files(self, filesystem, logging):
-        """Test that existing matching files are added to the slot."""
+    def test_check_and_add_existing_files_first_message(self, filesystem, logging):
+        """Test that existing matching files are added to the slot when the first message is received."""
         existing_files = [
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000007___-201611281100-__",
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000008___-201611281100-__",
             # A file that should not match
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG4__-MSG4________-VIS006___-000008___-201611281100-__"]
-        glob = MagicMock()
-        glob.return_value = existing_files
-        fs_ = MagicMock()
-        fs_.glob = glob
-        filesystem.return_value = fs_
-        mda = self.mda_msg0deg.copy()
-        fake_message = FakeMessage(mda)
-        message = Message(fake_message, self.msg0deg._patterns['msg'])
-        self.msg0deg._create_slot(message)
-        slot_str = str(mda["start_time"])
-        slot = self.msg0deg.slots[slot_str]
+        filesystem.return_value = _get_filesystem_return_value(existing_files)
+        message = _get_message_from_metadata_and_patterns(self.mda_msg0deg, self.msg0deg._patterns['msg'])
+        slot = _create_and_get_slot(self.msg0deg, message)
         self.msg0deg._config['check_existing_files_after_start'] = True
 
         self.msg0deg.check_and_add_existing_files(slot, message)
         assert call(logging.DEBUG) in logging.disable.mock_calls
         assert call(logging.NOTSET) in logging.disable.mock_calls
-        assert logging.disable.call_count == 2
         for fname in existing_files[:-1]:
             assert os.path.basename(fname) in slot._info['msg']['received_files']
         assert os.path.basename(existing_files[-1]) not in slot._info['msg']['received_files']
 
-        # For the next message(s) the existing files should not be handled
+    @patch("pytroll_collectors.segments.logging")
+    @patch("fsspec.filesystem")
+    def test_check_and_add_existing_files_not_first_message(self, filesystem, logging):
+        """Test that existing files are not checked/added after the first message."""
         existing_files = [
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000005___-201611281100-__",
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000006___-201611281100-__"]
-        glob.glob.return_value = existing_files
+        filesystem.return_value = _get_filesystem_return_value(existing_files)
+        message = _get_message_from_metadata_and_patterns(self.mda_msg0deg, self.msg0deg._patterns['msg'])
+        slot = _create_and_get_slot(self.msg0deg, message)
+        self.msg0deg._config['check_existing_files_after_start'] = True
+
+        self.msg0deg.check_and_add_existing_files(slot, message)
+        # Clear the received files and rerun, now the files should not be added
+        slot._info['msg']['received_files'] = {}
         self.msg0deg.check_and_add_existing_files(slot, message)
         for fname in existing_files[:-1]:
             assert os.path.basename(fname) not in slot._info['msg']['received_files']
-        assert logging.disable.call_count == 2
 
-        # Messages with other types than 'file' should be ignored
+    @patch("pytroll_collectors.segments.logging")
+    @patch("fsspec.filesystem")
+    def test_check_and_add_existing_files_ignore_collection_messages(self, filesystem, logging):
+        """Test that messages with other types than 'file' are ignored."""
+        message = _get_message_from_metadata_and_patterns(self.mda_msg0deg, self.msg0deg._patterns['msg'])
+        slot = _create_and_get_slot(self.msg0deg, message)
+        self.msg0deg._config['check_existing_files_after_start'] = True
         message.type = "collection"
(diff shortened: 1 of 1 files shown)


pnuu:
I created a new issue on documenting S3 configuration. It depends on completion of #110 and #114.

pnuu:
Hngh, I have somehow messed up parts of the PR and don't know where some of the code came from..

pnuu:
Ok, the results from `s3fs.S3FileSystem.glob()` don't include `'s3://'` in them. Now the collection works after a restart with usable URIs and I consider this completed (until review, at least :grin:).

As for the mysteriously appeared code fragment (https://github.com/pytroll/pytroll-collectors/pull/117/commits/53a8f74a038ff3297ae31814372c13134f7f2567#diff-76bd104754afbf31db6d4cf2bfa4f2df0dcb2fd57bbf7b9ca62b67be7cbeea8dL383), I have no idea where it came from. I just revered to the original (the version in `main`) for simplicity.

pytroll_collectors/segments.py
```diff
@@ -851,10 +851,35 @@ def _add_existing_files_to_slot(self, slot, fnames, message):
...
-    base_dir = os.path.dirname(path)
-
-    return glob.glob(os.path.join(base_dir, mask))
+    url_parts = urlparse(message.message_data["uri"])
+
+    if url_parts.scheme in ("", "file"):
+        base_dir = os.path.dirname(url_parts.path)
+        return glob.glob(os.path.join(base_dir, mask))
+    elif url_parts.scheme == "s3":
+        return _s3_glob(url_parts, mask)
```
mraspaud, line 860: how about using just fsspec here to have a more streamlined code? eg:
```python
    return _fs_glob(url_parts, mask)
```
where `_fs_glob` is `_s3_glob` with this line on top:
```python
    fs = fsspec.filesystem(url_parts.scheme, **fsoptions)
```

mraspaud (commented):
Thanks a lot for adding support for s3! I think we can make even more generic very easily.

pytroll_collectors/segments.py
pnuu: There are nothing called `fsoptions`.

mraspaud: yes, this is just a suggestion in case extra parameters need to be passed.

pnuu:
Finally remembered to document the S3 configuration. @mraspaud ok to merge?

doc/source/index.rst
```diff
@@ -272,6 +272,25 @@ Example yaml config:
 .. literalinclude:: ../../examples/segment_gatherer_msg_and_iodc.yaml_template
    :language: yaml
 
+If the collected segments are in an S3 object store, the
+``check_existing_files_after_start`` feature needs some additional
+configuration All the connection configurations and such are done
```
mraspaud, line 277:
```suggestion
configuration. All the connection configurations and such are done
```

doc/source/index.rst
```diff
@@ -272,6 +272,25 @@ Example yaml config:
...
+            "s3": {
+                "client_kwargs": {"endpoint_url": "https://s3.server.foo.com"},
+                "secret": "VERYBIGSECRET",
+                "key": "ACCESSKEY"
+            }
+        }
+
+
+.. _fsspec: https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration
+
```
mraspaud, line 292: oh cool, I didn't know that was possible!

pytroll_collectors/segments.py
mraspaud: I think we should use fsspec for every type of filesystem here, even local.

mraspaud (changes requested):
Still the pending question about the globing. Small comments otherwise.

pnuu:
With https://github.com/pytroll/pytroll-collectors/pull/117/commits/9af66a82b9aef392d5a4c3d8e2ba0636a0515a3c all `glob()` calls are done using `fsspec`.

pytroll_collectors/segments.py
pnuu: Done in https://github.com/pytroll/pytroll-collectors/commit/9af66a82b9aef392d5a4c3d8e2ba0636a0515a3c

adybbroe:
@mraspaud Do you have time to look at this one again?

pnuu:
@mraspaud one final look, please?

pytroll_collectors/tests/test_segments.py
```diff
@@ -607,14 +621,19 @@ def test_publish_service_name(self):
...
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000007___-201611281100-__",
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG3__-MSG3________-VIS006___-000008___-201611281100-__",
             # A file that should not match
             "/home/lahtinep/data/satellite/geo/msg/H-000-MSG4__-MSG4________-VIS006___-000008___-201611281100-__"]
+        glob = MagicMock()
+        glob.return_value = existing_files
+        fs_ = MagicMock()
+        fs_.glob = glob
+        filesystem.return_value = fs_
         glob.glob.return_value = existing_files
```
mraspaud, line 637: I'm not sure I understand why both `glob.return_value` and `glob.glob.return_value` need to be set to the same thing.

Also this test is pretty long, could it be split for clarity?

pytroll_collectors/tests/test_segments.py
```diff
@@ -651,6 +670,38 @@ def test_check_and_add_existing_files(self, glob, logging):
...
+        s3 = MagicMock()
+        s3.glob = glob
+        filesystem.return_value = s3
+        mda = self.mda_msg0deg_s3.copy()
+        fake_message = FakeMessage(mda)
+        message = Message(fake_message, self.msg0deg._patterns['msg'])
+        self.msg0deg._create_slot(message)
+        slot_str = str(mda["start_time"])
+        slot = self.msg0deg.slots[slot_str]
+        self.msg0deg._config['check_existing_files_after_start'] = True
```
mraspaud, line 694: This looks very similar to the previous test, can it be refactored to avoid duplication?

doc/source/index.rst
```diff
@@ -395,6 +395,25 @@ Example yaml config:
...
+        {
+            "s3": {
+                "client_kwargs": {"endpoint_url": "https://s3.server.foo.com"},
+                "secret": "VERYBIGSECRET",
+                "key": "ACCESSKEY"
+            }
+        }
+
+
+.. _fsspec: https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration
```
mraspaud, line 414: can the `.aws/credentials` file be used also?

pytroll_collectors/segments.py
```diff
@@ -851,10 +850,32 @@ def _add_existing_files_to_slot(self, slot, fnames, message):
...
+            '',
+            ''
+        )
+    )
+
+    fs_ = fsspec.filesystem(url_parts.scheme)
+    files = fs_.glob(pattern)
+    # There might be no scheme in the returned filenames, so add it if scheme is defined
+    if url_parts.scheme:
+        files = [url_parts.scheme + '://' + f for f in files if not f.startswith(
```
mraspaud, line 876: Wouldn't `urlunparse` be preferable here?

mraspaud (commented):
I just have a few questions and style comments left, otherwise looks good!

pytroll_collectors/tests/test_segments.py
pnuu: The `glob.glob` one is just a mistake, I tend to do `import glob` and `glob.glob(pattern)` so I've thought I'm using `glob.glob` here even if I'm not. Removing that. and having a look at restructuring the test. It should at least be 3 separate tests.

doc/source/index.rst
pnuu: No idea. The above is what `fsspec` says it can use.

pytroll_collectors/segments.py
pnuu: Using `urlunparse()` the one-liner would be:
```python
    if url_parts.scheme:
        files_with_scheme = []
        for f in files:
            parts = list(urlparse(f))
            parts[0] = url_parts.scheme
            # Netloc is empty, so we need to add the scheme slashes manually
            parts[1] = "//"
            files_with_scheme.append(urlunparse(parts))
        files = files_with_scheme
    return files
```
Not an improvement in my opinion.

pytroll_collectors/tests/test_segments.py
pnuu: Restructured in b0b769843210f35bc22d952d4fbf9d15de99d2bb

pytroll_collectors/tests/test_segments.py
pnuu: Restructured in b0b769843210f35bc22d952d4fbf9d15de99d2bb

pytroll_collectors/segments.py
mraspaud: indeed, it doesn't look better. It's just the ` + "://" + ` I wasn't so happy about, but we can leave it be.

pnuu: Oh, and this alternate version would break if there were a `netloc` in the URI.

mraspaud (approved):
LGTM

PR #118, opened 2022-09-09
Ensure files in S3 keep their full URIs in messages

pnuu:
This PR makes sure that segments collected from S3 keep their full URIs. Previously the scheme and bucket name were stripped away.


mraspaud:
Should be just pass through everything and let downstream applications do the stripping?

pnuu:
Maybe. This just keeps the backwards compatibility.

mraspaud:
ok, but for consistency then? the geographic gatherer lets the uri pass through... trollflow/satpy can handle these anyway, right?

pnuu:
Good point. Now all incoming URIs are passed as-is.

mraspaud (approved):
LGTM!

PR #124, opened 2022-12-13
Use tolerance also when using start_time in message_keys 

ninahakansson:
When using start_time in the message_keys the time tolerance is ignored. Matching files within time tolerance are ignored in the is_relevant function and not included in any slot.

### DIFF
--- a/pytroll_collectors/segments.py
+++ b/pytroll_collectors/segments.py
@@ -42,6 +42,7 @@ from collections import OrderedDict
 from enum import Enum
 import glob
 import os
+from datetime import datetime
 
 import trollsift
 from posttroll import message as pmessage
@@ -168,6 +169,7 @@ class Message:
         self.metadata = pattern.parser.parse(self.message_data)
         self._time_name = self.pattern.time_name
         self._adjust_time_by_flooring()
+        self.slot_time = None  # Might differ by tolerance (s) from time_name
 
     @property
     def id_time(self):
@@ -406,7 +408,18 @@ class Slot:
         if mask in slot_pattern['received_files']:
             logger.debug("File already received")
             should_be_added = False
-        if mask not in slot_pattern['all_files']:
+        if mask in slot_pattern['all_files']:
+            pass
+        elif message.slot_time is not None:
+            # File probably within tolerance of existing slot
+            metadata_new = metadata.copy()
+            metadata_new[message._time_name] = message.slot_time  # Use slot time insted of file_time
+            mask_temp = message.pattern.parser.globify(metadata_new)
+            if mask_temp not in slot_pattern['all_files']:
+                logger.debug("%s not in %s", mask, slot_pattern['all_files'])
+                logger.debug("%s not in %s", mask_temp, slot_pattern['all_files'])
+                should_be_added = False
+        else:
             logger.debug("%s not in %s", mask, slot_pattern['all_files'])
             should_be_added = False
         return mask, should_be_added
@@ -754,7 +767,10 @@ class SegmentGatherer(object):
                 return
 
         slot_time = self._find_time_slot(message.id_time)
-
+        try:
+            message.slot_time = datetime.strptime(slot_time, "%Y-%m-%d %H:%M:%S.%f")
+        except ValueError:
+            message.slot_time = datetime.strptime(slot_time, "%Y-%m-%d %H:%M:%S")
         # Init metadata etc if this is the first file
         if slot_time not in self.slots:
             slot = self._create_slot(message)
--- a/pytroll_collectors/tests/test_segments.py
+++ b/pytroll_collectors/tests/test_segments.py
@@ -28,7 +28,6 @@ import os
 import os.path
 import unittest
 from unittest.mock import patch, MagicMock, call
-
 import pytest
 
 import posttroll.message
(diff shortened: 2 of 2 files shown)


pytroll_collectors/segments.py
```diff
@@ -406,7 +408,18 @@ def is_relevant(self, message):
         if mask in slot_pattern['received_files']:
             logger.debug("File already received")
             should_be_added = False
-        if mask not in slot_pattern['all_files']:
+        if mask in slot_pattern['all_files']:
+            pass
```
adybbroe, line 412: It maybe me, but I get a bit confused using a pass statement like this, wouldn't it be more clear to just do a 
`return mask, should_be_added` here directly instead!?

pytroll_collectors/segments.py
```diff
@@ -406,7 +408,18 @@ def is_relevant(self, message):
...
+        elif message.slot_time is not None:
+            # File probably within tolerance of existing slot
+            metadata_new = metadata.copy()
+            metadata_new[message._time_name] = message.slot_time  # Use slot time insted of file_time
+            mask_temp = message.pattern.parser.globify(metadata_new)
+            if mask_temp not in slot_pattern['all_files']:
+                logger.debug("%s not in %s", mask, slot_pattern['all_files'])
+                logger.debug("%s not in %s", mask_temp, slot_pattern['all_files'])
+                should_be_added = False
+        else:
```
adybbroe, line 422: And then this could be replaced by a function call like:
`should_be_added = check_if_file_is_within_tolerance_of_existing_slot(message, metadata, mask)`
And you could then skip the comments! Adding comments in code is usually a sign that the code doesn't read for itself.

adybbroe (changes requested):
Thanks. Looks good in general. Some few comments - some functions are getting a bit long and would benefit from a bit of refactoring

adybbroe:
@ninahakansson Are we ready to finish this one?

ninahakansson:
My last commit was an update to a test that makes the test fail. On purpose. The code should be fixed so that the test does not fail.

pytroll_collectors/segments.py
adybbroe: @pnuu or @mraspaud may have different opinions?

pytroll_collectors/tests/test_segments.py
```diff
@@ -988,13 +1016,27 @@ def test_mismatching_files_generate_multiple_slots(self):
...
+        viirs_msg = Message_p(rawstr=viirs_message)
+        pps_msg = FakeMessage(new_pps_message_data_within_time_tolerance, message_type='dataset', subject='/foo/pps')
+
+        self.collection_gatherer.process(viirs_msg)
+        self.collection_gatherer.process(pps_msg)
+
+        assert len(self.collection_gatherer.slots) == 1
+        slot = self.collection_gatherer.slots['2020-10-13 05:17:21.200000']
+        assert slot.output_metadata['collection']['pps']['dataset'] != []
+        # assert slot.get_status() == Status.SLOT_READY
```
adybbroe, line 1038: What about this comment? Something you had in mind implementing?
And the last check only tests that the variable is not an empty list. Can this be more explicit?

ninahakansson: I fixed the time tolerance half-way. The status is not set to ready even though the correct files are found. Adding the  assert slot.get_status() == Status.SLOT_READY will make the test fail as step 1 in TDD. But to fix it I think is more difficult. So we decided to make an issue instead https://github.com/pytroll/pytroll-collectors/issues/129 so that this PR could be merged. I left the line in the code for later.

pytroll_collectors/segments.py
pnuu: I don't have a strong opinion. The other branches do checks and modifies `should_be_added` accordingly and don't do a return.

pytroll_collectors/tests/test_segments.py
```diff
@@ -988,13 +1016,36 @@ def test_mismatching_files_generate_multiple_slots(self):
...
+            {'uri': '/san1/polar_out/direct_readout/lvl2/S_NWC_CMA_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc',  # noqa
+             'uid': 'S_NWC_CMA_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc'},
+            {'uri': '/san1/polar_out/direct_readout/lvl2/S_NWC_CTTH_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc',  # noqa
+             'uid': 'S_NWC_CTTH_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc'},
+            {'uri': '/san1/polar_out/direct_readout/lvl2/S_NWC_CT_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc',  # noqa
+             'uid': 'S_NWC_CT_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc'},
+            {'uri': '/san1/polar_out/direct_readout/lvl2/S_NWC_CPP_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc',  # noqa
+             'uid': 'S_NWC_CPP_noaa20_14587_20200911T1205084Z_20200911T1206312Z.nc'}]
+        assert slot.output_metadata['collection']['pps']['dataset'] == expected_collection
+        # assert slot.get_status() == Status.SLOT_READY
```
mraspaud, line 1047: Please remove the commented code.
```suggestion
```

pnuu:
I fixed the merge conflict.

pnuu:
Hmm, there are now 48 failing tests. Must be something that came from `main`. I'll have a look.

pnuu:
For some reason the change done to `Message.__init__()` when flooring the time wasn't taken from `main`. Now the tests pass.

@adybbroe @mraspaud do you have still something that is needed in the unresolved comments?

pytroll_collectors/segments.py
mraspaud: I agree with @adybbroe here, this could be refactored a bit.

pytroll_collectors/segments.py
mraspaud: If we can return immediately it's best to do it, yes.

mraspaud (changes requested):
Just some style comments

PR #125, opened 2022-12-19
Change publish topic to start with the instrument name

adybbroe:
This PR introduce the instrument name to the publish topic so e.g. that instead of `/RDR/0...` it would be `/atms/RDR/0...`

### DIFF
--- a/pytroll_collectors/scisys.py
+++ b/pytroll_collectors/scisys.py
@@ -493,7 +493,7 @@ class GMCSubscriber(object):
         self.loop = False
 
 
-def _get_subject_from_msg2send(to_send, station, environment, topic_postfix):
+def _get_subject_from_msg2send(to_send, station, env, topic_postfix):
     """Get the publish topic from the message to be send."""
     if topic_postfix is not None:
         subject = "/".join(("", to_send['sensor'], to_send['format'],
@@ -502,7 +502,7 @@ def _get_subject_from_msg2send(to_send, station, environment, topic_postfix):
     else:
         subject = "/".join(("", to_send['sensor'], to_send['format'],
                             to_send['data_processing_level'],
-                            station, environment,
+                            station, env,
                             "polar", "direct_readout"))
     return subject
 
@@ -534,7 +534,7 @@ def receive_from_zmq(host, port, station, environment, excluded_platforms,
             logger.debug("to_send: %s", str(to_send))
             if to_send is None:
                 continue
-            subject = _get_subject_from_msg2send(to_send, topic_postfix)
+            subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
             logger.debug("Subject: %s", str(subject))
             msg = Message(subject,
                           "file",
--- a/pytroll_collectors/scisys.py
+++ b/pytroll_collectors/scisys.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2012 - 2021 Pytroll developers
+# Copyright (c) 2012 - 2022 Pytroll developers
 #
 # Author(s):
 #
@@ -140,9 +140,9 @@ class PassRecorder(dict):
         """Get a pass."""
         utctime, satellite = key
         for (rectime, recsat), val in self.iter():
-            if(recsat == satellite and
-               (abs(rectime - utctime)).seconds < 30 * 60 and
-               (abs(rectime - utctime)).days == 0):
+            if (recsat == satellite and
+                (abs(rectime - utctime)).seconds < 30 * 60 and
+                    (abs(rectime - utctime)).days == 0):
                 return val
         return default
 
@@ -493,6 +493,20 @@ class GMCSubscriber(object):
         self.loop = False
 
 
+def _get_subject_from_msg2send(to_send, station, environment, topic_postfix):
+    """Get the publish topic from the message to be send."""
+    if topic_postfix is not None:
+        subject = "/".join(("", to_send['sensor'], to_send['format'],
+                            to_send['data_processing_level'],
+                            topic_postfix))
+    else:
+        subject = "/".join(("", to_send['sensor'], to_send['format'],
+                            to_send['data_processing_level'],
+                            station, environment,
+                            "polar", "direct_readout"))
+    return subject
+
+
 def receive_from_zmq(host, port, station, environment, excluded_platforms,
                      target_server, ftp_prefix, topic_postfix,
                      publish_port=0, nameservers=None, days=1):
@@ -520,15 +534,7 @@ def receive_from_zmq(host, port, station, environment, excluded_platforms,
             logger.debug("to_send: %s", str(to_send))
             if to_send is None:
                 continue
-            if topic_postfix is not None:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   topic_postfix))
-            else:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   station, environment,
-                                   "polar", "direct_readout"))
+            subject = _get_subject_from_msg2send(to_send, topic_postfix)
             logger.debug("Subject: %s", str(subject))
             msg = Message(subject,
                           "file",
--- a/pytroll_collectors/scisys.py
+++ b/pytroll_collectors/scisys.py
@@ -495,12 +495,16 @@ class GMCSubscriber(object):
 
 def _get_subject_from_msg2send(to_send, station, env, topic_postfix):
     """Get the publish topic from the message to be send."""
+    if isinstance(to_send['sensor'], str):
+        prefix = '/' + to_send['sensor'].replace('/', '-')
+    else:
+        prefix = ''
     if topic_postfix is not None:
-        subject = "/".join(("", to_send['sensor'], to_send['format'],
+        subject = "/".join((prefix, to_send['format'],
                             to_send['data_processing_level'],
                             topic_postfix))
     else:
-        subject = "/".join(("", to_send['sensor'], to_send['format'],
+        subject = "/".join((prefix, to_send['format'],
                             to_send['data_processing_level'],
                             station, env,
                             "polar", "direct_readout"))
--- a/pytroll_collectors/scisys.py
+++ b/pytroll_collectors/scisys.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2012 - 2022 Pytroll developers
+# Copyright (c) 2012 - 2023 Pytroll developers
 #
 # Author(s):
 #
@@ -493,7 +493,7 @@ class GMCSubscriber(object):
         self.loop = False
 
 
-def _get_subject_from_msg2send(to_send, station, env, topic_postfix):
+def get_subject_from_msg2send(to_send, station, env, topic_postfix):
     """Get the publish topic from the message to be send."""
     if isinstance(to_send['sensor'], str):
         prefix = '/' + to_send['sensor'].replace('/', '-')
@@ -538,7 +538,7 @@ def receive_from_zmq(host, port, station, environment, excluded_platforms,
             logger.debug("to_send: %s", str(to_send))
             if to_send is None:
                 continue
-            subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
+            subject = get_subject_from_msg2send(to_send, station, environment, topic_postfix)
             logger.debug("Subject: %s", str(subject))
             msg = Message(subject,
                           "file",
--- a/pytroll_collectors/tests/test_scisys.py
+++ b/pytroll_collectors/tests/test_scisys.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2013 - 2021 Pytroll developers
+# Copyright (c) 2013 - 2022 Pytroll developers
 #
 # Author(s):
 #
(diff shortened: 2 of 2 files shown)


pytroll_collectors/scisys.py
```diff
@@ -520,15 +534,7 @@ def receive_from_zmq(host, port, station, environment, excluded_platforms,
...
-            if topic_postfix is not None:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   topic_postfix))
-            else:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   station, environment,
-                                   "polar", "direct_readout"))
+            subject = _get_subject_from_msg2send(to_send, topic_postfix)
```
pnuu, line 537: According to Codecov, this line where the subject forming (= the change made here) happens isn't covered by the tests.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -216,14 +217,51 @@ def test_reception(self):
...
+    filename = os.path.join('/tmp', VIIRS['uid'])
+    touch(filename)
+    string = TwoMetMessage(INPUT_DISPATCH_VIIRS)
+    to_send = msg_rec.receive(string)
+    os.remove(filename)
 
+    topic_postfix = None
+    station = 'nrk'
+    environment = 'dev'
+    subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
```
pnuu, line 233: The tests should call the public interface (`receive_from_zmq()`).

pytroll_collectors/tests/test_scisys.py
```diff
@@ -216,14 +217,51 @@ def test_reception(self):
...
+def test_get_subject_from_msg2send_viirs():
+    """Test the get the subject from the message being send - viirs data."""
+    msg_rec = MessageReceiver("nimbus")
 
-    return mysuite
+    filename = os.path.join('/tmp', VIIRS['uid'])
+    touch(filename)
+    string = TwoMetMessage(INPUT_DISPATCH_VIIRS)
+    to_send = msg_rec.receive(string)
+    os.remove(filename)
```
pnuu, line 228: The file will be left behind if there's an exception here. Instead I'd do something like

```python
with tempfile.TemporaryDirectory as tmp:
    filename = os.path.join(tmp, VIIRS['uid'])
    touch(filename)
    string = ...
    to_send = ...
```
The context manager will remove also the created temporary file.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -216,14 +217,51 @@ def test_reception(self):
...
-if __name__ == '__main__':
-    unittest.main()
+    assert subject == "/viirs/RDR/0/nrk/dev/polar/direct_readout"
+
+    station = 'nrk'
+    environment = 'dev'
+    topic_postfix = 'my_topic'
+    subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
+
+    assert subject == "/viirs/RDR/0/my_topic"
```
pnuu, line 242: This is another test case, so should be separate to the previous one.

pytroll_collectors/tests/test_scisys.py
pnuu: Oh, and making this a function that takes the UID as argument and returns `to_send` would make it reusable (see below).

pytroll_collectors/tests/test_scisys.py
```diff
@@ -216,14 +217,51 @@ def test_reception(self):
...
+
+def test_get_subject_from_msg2send_atms():
+    """Test get the subject from the message being send - atms data."""
+    msg_rec = MessageReceiver("nimbus")
+
+    filename = os.path.join('/tmp', ATMS['uid'])
+    touch(filename)
+    string = TwoMetMessage(INPUT_DISPATCH_ATMS)  # input_dispatch_atms)
+    to_send = msg_rec.receive(string)
+    os.remove(filename)
```
pnuu, line 253: Get `to_send` using the new function from earlier.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -216,14 +217,51 @@ def test_reception(self):
...
+    subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
+
+    assert subject == "/atms/RDR/0/nrk/dev/polar/direct_readout"
+
+    station = 'nrk'
+    environment = 'dev'
+    topic_postfix = ''
+    subject = _get_subject_from_msg2send(to_send, station, environment, topic_postfix)
+
+    assert subject == "/atms/RDR/0/"
```
pnuu, line 267: These are two separate testcases, and should use the public interface.

pnuu (commented):
The main question for me is: does the topic need to be formed in the code, or could it be user-definable? This change is also backwards incompatible and will require changes downstream.

Although, as we are not using scisys receiver (as far as I remember), I'm not too worried how the usage changes :-)

Some comments/suggestions inline.

pytroll_collectors/tests/test_scisys.py
adybbroe: Done

pytroll_collectors/tests/test_scisys.py
pnuu: This is still using the private interface (= implementation).

pnuu (commented):
One unhandled comment in the tests.

And how about the user-configurable topic?

adybbroe:
> One unhandled comment in the tests.
> 
> And how about the user-configurable topic?

Sorry, what do you mean "And how about the user-configurable topic"?

And concerning testing the private interface, what is the issue? What do you suggest?
I re factored out an untested part in `receive_from_zmq` to test it, but I think it should still be a private function. Testing the entire `receive_from_zmq` as is, is beyond what I am capable of.

pnuu:
> > One unhandled comment in the tests.
> > And how about the user-configurable topic?
> 
> Sorry, what do you mean "And how about the user-configurable topic"?

I can't find my previous comment, but should it be possible for the user to configure the topic instead of forcing the topic to be what scisys receiver is now setting it to?

> And concerning testing the private interface, what is the issue?

Private interface indicates that it's an implementation detail and might change.

> What do you suggest?

What we normally do is to create a test that calls the public interface and check that it produces the expected result with different settings.

adybbroe:
I have converted the private function to a public one.
I suggest stopping the re-factoring here. This part of the project is in need for some more care and refactoring and increased test coverage, but I really don't think this belongs here. I hope you can both agree @mraspaud and @pnuu ?

pytroll_collectors/scisys.py
```diff
@@ -520,15 +538,7 @@ def receive_from_zmq(host, port, station, environment, excluded_platforms,
...
-            if topic_postfix is not None:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   topic_postfix))
-            else:
-                subject = "/".join(("", to_send['format'],
-                                   to_send['data_processing_level'],
-                                   station, environment,
-                                   "polar", "direct_readout"))
+            subject = get_subject_from_msg2send(to_send, station, environment, topic_postfix)
```
pnuu, line 541: This is still untested. The only case now tested is when `to_send` is `None`.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
+
+        with tempfile.TemporaryDirectory() as tmp:
+            filename = os.path.join(tmp, VIIRS['uid'])
+            touch(filename)
+            string = TwoMetMessage(create_input_dispatch_viirs_msg(tmp))
+            to_send = msg_rec.receive(string)
+            viirs_cpy = deepcopy(VIIRS)
+            viirs_cpy.pop('uri')
+            to_send.pop('uri')
+            TestCase().assertDictEqual(to_send, viirs_cpy)
```
pnuu, line 207: This structure is repeated a lot, so should be refactored.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
+            os.remove(filename)
```
pnuu, line 208: This is not needed, the context manager will remove the whole `tmp` tree when exited.

pytroll_collectors/tests/test_scisys.py
pnuu: And each of these `with` branches should be a separate test case.

pnuu (commented):
Few more new comments.

Sorry if these were not originally part of this PR, but the review system shows them changed so added the comments. It's hard to keep track what's new and what has been adjusted for some other reason.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -63,69 +77,113 @@
...
 
-msg_terra = {"platform_name": "EOS-Terra", "uri":
-             "ssh://{hostname}/tmp/P0420064AAAAAAAAAAAAAA14303204950001.PDS".format(hostname=hostname), "format": "PDS",
-             "start_time": datetime.datetime(2014, 10, 30, 20, 49, 50),
-             "data_processing_level": "0", "orbit_number": 79082, "uid":
-             "P0420064AAAAAAAAAAAAAA14303204950001.PDS",
-             "sensor": "modis",
-             "end_time": datetime.datetime(2014, 10, 30, 21, 3, 50),
-             "type": "binary", 'variant': 'DR'}
+def create_fildis_terra(dirname):
```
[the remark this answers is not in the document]
adybbroe: FILDIS comes from the GMCServer I believe, so a SCISYS/CGI naming.

pytroll_collectors/scisys.py
adybbroe: Well, yes, correct, the whole `with Publish`was untested. But now I introduced (re-factored) a function to create the adequate subject (`get_subject_from_msg2send`) that I believe is tested for various use-cases, not only when `to_send` is `None`. Or am I mistaken?

pytroll_collectors/tests/test_scisys.py
```diff
@@ -63,69 +77,113 @@
...
+                 'variant': 'DR'}
+
+
 def touch(fname):
     """Create an empty file."""
     open(fname, 'a').close()
 
 
-class ScisysReceiverTest(unittest.TestCase):
+class TestScisysReceiver:
```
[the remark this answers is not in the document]
adybbroe: Fixed, thanks!

pytroll_collectors/tests/test_scisys.py
adybbroe: Fixed. This can be further improved I am sure, but now it will be easier next time one is in for a re-feactoring I think.

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
+            os.remove(filename)
+
+
+@pytest.mark.parametrize("sensor, sensor_name", [(VIIRS, 'viirs'),
+                                                 (ATMS, 'atms')])
+def test_get_subject_from_msg2send_postfix_topic_is_none(sensor, sensor_name):
+    """Test the get the subject from the message being send."""
+    msg_rec = MessageReceiver("nimbus")
+
+    filename = os.path.join('/tmp', sensor['uid'])
```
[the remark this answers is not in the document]
adybbroe: Should be fixed now!

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
+    assert subject == "/{sensor}/RDR/0/nrk/dev/polar/direct_readout".format(sensor=sensor_name)
+
+
+@pytest.mark.parametrize("sensor, sensor_name", [(VIIRS, 'viirs'),
+                                                 (ATMS, 'atms')])
+def test_get_subject_from_msg2send_with_postfix_topic(sensor, sensor_name):
+    """Test the get the subject from the message being send."""
+    msg_rec = MessageReceiver("nimbus")
+
+    filename = os.path.join('/tmp', sensor['uid'])
```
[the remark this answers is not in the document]
adybbroe: Should be fixed now!

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
+    subject = get_subject_from_msg2send(to_send, station, environment, topic_postfix)
+
+    assert subject == "/{sensor}/RDR/0/my_topic".format(sensor=sensor_name)
+
+
+def test_get_subject_from_msg2send_empty_postfix_topic():
+    """Test get the subject from the message being send - empty postfix topic."""
+    msg_rec = MessageReceiver("nimbus")
+
+    filename = os.path.join('/tmp', ATMS['uid'])
```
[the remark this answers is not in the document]
adybbroe: Again, hope you agree this is okay now?

pytroll_collectors/tests/test_scisys.py
```diff
@@ -136,94 +194,159 @@ def test_reception(self):
...
-    """Test suite for test_scisys."""
-    loader = unittest.TestLoader()
-    mysuite = unittest.TestSuite()
-    mysuite.addTest(loader.loadTestsFromTestCase(ScisysReceiverTest))
+def test_get_subject_from_msg2send_avhrr3():
+    """Test get the subject from the message being send - avhrr data."""
+    msg_rec = MessageReceiver("nimbus")
 
-    return mysuite
+    filename = os.path.join('/tmp', create_msg_n19('/tmp')['uid'])
```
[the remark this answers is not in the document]
adybbroe: Smae as above. Believe it is fixed.

pytroll_collectors/scisys.py
pnuu: Codecov is telling that `scisys.py` line 542 is untested. I guess the new tests call `get_subject_from_msg2send()` and thus not going this route and so the line is not covered. The rest of the `receive_from_zmq()` function, including the publishing, is also uncovered due to this.

pnuu (commented):
Just one comment. Not critical if you think it's good enough to test the lower level `get_subject_from_msg2send()`.

And again, I think you're the only user for this so it's all up to you I think :-)

adybbroe:
Thanks @pnuu Yes, we are likely the only ones. And we will probably not have this CGI based system in a year from now, and as you said this code block was not tested before either. So anyhow this is an incremental improvement.
Maybe just let @mraspaud see if he agress before merging?

PR #130, opened 2023-02-14
Add a mapscript recorder

mraspaud:
This PR adds a recorder that records incoming messages into a file for use by eg map scripts.


mraspaud:
This is not needed anymore, closing.

PR #132, opened 2023-02-21
Fix group_by_minutes to include composed filenames

mraspaud:
FCI files where not being gathered correctly. This PR fixes it.


pnuu (approved):
LGTM

PR #133, opened 2023-03-22
Make it possible to tell segment gatherer that all the files are local

pnuu:
With this PR it is possible to set a configuration option (`all_files_are_local = True`) that tells segment gatherer all the files to be local independent of the possible transport scheme (for example `scp://`) or host (`foo.bar.com`) in the received message metadata URI.

### DIFF
--- a/pytroll_collectors/segments.py
+++ b/pytroll_collectors/segments.py
@@ -158,10 +158,12 @@ class MessageParser(Parser):
 class Message:
     """A message object."""
 
-    def __init__(self, posttroll_message, pattern):
+    def __init__(self, posttroll_message, pattern, drop_scheme=False):
         """Set up the message."""
         self.pattern = pattern
-        self.message_data = posttroll_message.data.copy()
+        self._drop_scheme = drop_scheme
+        posttroll_message = self._check_scheme(posttroll_message)
+        self.message_data = posttroll_message.data
         self.type = posttroll_message.type
         self._posttroll_message = posttroll_message
         self.metadata = pattern.parser.parse(self.message_data)
@@ -190,6 +192,24 @@ class Message:
         time_item = dt.datetime(time_item.year, time_item.month,
                                 time_item.day, time_item.hour, floor_minutes, 0)
         metadata[time_name] = time_item
+
+    def _check_scheme(self, posttroll_message):
+        message_data = posttroll_message.data.copy()
+        if self._drop_scheme:
+            url_parts = urlparse(message_data['uri'])
+            uri = urlunparse(
+                (
+                    '',
+                    '',
+                    url_parts.path,
+                    '',
+                    '',
+                    ''
+                )
+            )
+            message_data['uri'] = uri
+            posttroll_message.data = message_data
+        return posttroll_message
 
     @property
     def filtered_metadata(self):
@@ -782,7 +802,8 @@ class SegmentGatherer(object):
         for pattern in self._patterns.values():
             try:
                 if pattern.parser.matches(msg):
-                    return Message(msg, pattern)
+                    drop_scheme = self._config.get('all_files_are_local', False)
+                    return Message(msg, pattern, drop_scheme=drop_scheme)
             except KeyError as err:
                 logger.debug("No key %s in message.", str(err))
         raise TypeError
@@ -997,6 +1018,11 @@ def ini_to_dict(fname, section):
         conf['check_existing_files_after_start'] = config.getboolean(section, "check_existing_files_after_start")
     except (NoOptionError, ValueError):
         conf['check_existing_files_after_start'] = False
+
+    try:
+        conf['all_files_are_local'] = config.getboolean(section, "all_files_are_local")
+    except (NoOptionError, ValueError):
+        conf['all_files_are_local'] = False
 
     return conf
 
--- a/pytroll_collectors/segments.py
+++ b/pytroll_collectors/segments.py
@@ -162,7 +162,7 @@ class Message:
         """Set up the message."""
         self.pattern = pattern
         self._drop_scheme = drop_scheme
-        posttroll_message = self._check_scheme(posttroll_message)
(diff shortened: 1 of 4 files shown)


pnuu:
Did a live test and the schema and host were stripped as intended.

pytroll_collectors/tests/test_segments.py
```diff
@@ -77,15 +76,12 @@ def __init__(self, data, message_type='file', subject='/foo/viirs'):
...
+class TestSegmentGatherer:
     """Tests for the segment gatherer."""
 
     @pytest.fixture(autouse=True)
-    def inject_fixtures(self, caplog):
-        """Inject fixtures."""
-        self._caplog = caplog
-
-    def setUp(self):
+    def setup(self):
```
mraspaud, line 84: I think if you call this `setup_method`, then the fixture autouse command above could be removed.

pytroll_collectors/tests/test_segments.py
```diff
@@ -77,15 +76,12 @@ def __init__(self, data, message_type='file', subject='/foo/viirs'):
         self.subject = subject
 
 
-class TestSegmentGatherer(unittest.TestCase):
+@pytest.mark.usefixtures("caplog")
```
mraspaud, line 79: I'm surprised this is needed? Could you explain why?

pytroll_collectors/tests/test_segments.py
```diff
@@ -329,134 +325,152 @@ def test_get_collection_status(self):
...
+        """Test processing message with scheme when config says files are local and existing files are checked."""
+        mda = self.mda_msg0deg.copy()
+        mda['uri'] = 'SCHEME://host.foo.bar' + mda['uri']
+        msg = FakeMessage(mda)
+        col = SegmentGatherer(CONFIG_SINGLE)
+        # Without forcing all files to be local the processing would fail with
+        # "ValueError: Protocol not known: scheme"
+        col._config['all_files_are_local'] = True
+        col._config['check_existing_files_after_start'] = True
+        col.process(msg)
```
mraspaud, line 472: Do we need to add an assert here for something?

pytroll_collectors/tests/test_segments.py
```diff
@@ -1319,22 +1329,23 @@ def test_floor_time_2_minutes_with_seconds(self):
...
+            assert self.himawari_ini_message.metadata['start_time'].minute == 29
 
 
-class TestFlooringMultiplePatterns(unittest.TestCase):
+class TestFlooringMultiplePatterns:
     """Test flooring."""
 
-    def setUp(self):
+    @pytest.fixture(autouse=True)
+    def setup(self):
```
mraspaud, line 1348: using `setup_method` should allow getting rid of the autouse call

pytroll_collectors/tests/test_segments.py
```diff
@@ -1275,10 +1284,11 @@ def test_get_unique_id_from_dataset_message(self):
         assert message.uid().startswith('NOAA-20_2020-10-13 05:17:21')
 
 
-class TestFlooring(unittest.TestCase):
+class TestFlooring:
     """Test flooring."""
 
-    def setUp(self):
+    @pytest.fixture(autouse=True)
+    def setup(self):
```
mraspaud, line 1291:
```suggestion
    def setup_method(self):
```

pytroll_collectors/tests/test_segments.py
```diff
@@ -1032,10 +1044,12 @@ def _assert_listener_container(listener_container, listener_nameserver):
...
 
 
-class TestSegmentGathererCollections(unittest.TestCase):
+@pytest.mark.usefixtures("caplog")
+class TestSegmentGathererCollections:
     """Test collections gathering."""
 
-    def setUp(self):
+    @pytest.fixture(autouse=True)
+    def setup(self):
```
mraspaud, line 1052:
```suggestion
    def setup_method(self):
```

pytroll_collectors/segments.py
```diff
@@ -191,6 +193,24 @@ def _adjust_time_by_flooring(self, metadata, group_by_minutes, time_name):
                                 time_item.day, time_item.hour, floor_minutes, 0)
         metadata[time_name] = time_item
 
+    def _check_scheme(self, posttroll_message):
```
mraspaud, line 196: just a naming comment here, how about `_handle_scheme` instead? For me it conveys better that the message might be changed.

mraspaud (commented):
Thank you for this PR, and for taking the time to refactor the tests to use pytest exclusively. Some comments inline, but I don't see any problem in general.
Also, the documentation needs to be updated I think https://github.com/pytroll/pytroll-collectors/blob/main/doc/source/index.rst#segment_gatherer

pytroll_collectors/tests/test_segments.py
pnuu: Without the handling added in this PR the test will crash, so I thought just that the test passes would be enough. If you have an idea what else to check, I'm all ears.

pytroll_collectors/tests/test_segments.py
pnuu: Seems so. Fixed in https://github.com/pytroll/pytroll-collectors/pull/133/commits/e727700a21ccfaeb0d12bd4b686c062b271fd8a5

pytroll_collectors/tests/test_segments.py
pnuu: Fixed in https://github.com/pytroll/pytroll-collectors/pull/133/commits/15ea90e4cf7c333c10f204bd1d3954ee736e1fca

pytroll_collectors/tests/test_segments.py
mraspaud: that's all good :)

pnuu:
The new option is now also in the documentation.

PR #135, opened 2023-04-25
Fix calling s3.ls() to get also new files for the daemon version

pnuu:
The `s3.ls()` call was missing `refresh=True` kwarg that is required by the daemon version of the S3Stalker. Without it, the call results are taken from cache and none of the new files are returned.


TAlonglong (approved):
LGTM

adybbroe (approved):
LGTM, Many thanks!

PR #136, opened 2023-04-26
Drop S3 credentials from messages in S3Stalker (+ daemon)

pnuu:
The messages published by S3Stalker(daemon) had all the credentials published in the messages. ~This PR drops the whole `filesystem` part of the message, which needs to be configured at the receiving side in any case.~

Update: only drop known S3 credentials.


pnuu:
Apparently the fileystem stuff was expected to be in the messages, so lets just drop the S3 credentials we know of.

TAlonglong (approved):
LGTM

PR #139, opened 2023-05-22
Create monitored directory if does not exist

tecnavia-dev:
Create monitored directory, if does not exist


mraspaud (changes requested):
Thanks for this PR. Could you write a small test for this?

pytroll_collectors/trollstalker.py
```diff
@@ -262,11 +262,22 @@ def create_notifier(topic, instrument, posttroll_port, filepattern,
 
     # Add directories and event masks to watch manager
     for monitored_dir in monitored_dirs:
+        # Create directory, if it does not exist
```
mraspaud, line 265: This comment is not needed...

mraspaud (commented):
Thanks for adding this functionality, some comments inline

pytroll_collectors/trollstalker.py
tecnavia-dev: Ok

pytroll_collectors/trollstalker.py
```diff
@@ -262,11 +262,22 @@ def create_notifier(topic, instrument, posttroll_port, filepattern,
...
     return notifier
 
 
+def check_if_monitored_dir_exist(directory):
+    """Create directory, if it does not exist."""
+    try:
+        os.makedirs(directory, exist_ok=True)
+    except PermissionError as err:
+        logger.error(f"Can't create monitored directory: {err}")
+        raise
```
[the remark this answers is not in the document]
tecnavia-dev: I don't think, we need it.
I can move it, so It'll be cleaner.

tecnavia-dev:
Do we need to make any further changes?

mraspaud (approved):
LGTM

PR #141, opened 2023-05-24
Add tests and refactor trollstalker a bit

mraspaud:
This PR adds a simple functionality test for trollstalker.
At the same time the main functionality is moved out of the bin directory into the pytroll_collector package.

Another test is commented out for now, as this will be used in the next PR.


pytroll_collectors/trollstalker.py
```diff
@@ -0,0 +1,459 @@
...
+    for key in config:
+        if 'var_' in key:
+            new_key = key.replace('var_', '')
+            var = config[key]
+            variables[new_key] = var
+    return variables
+
+
+def main(command_args=None):
+    """Run Trollstakler.
```
pnuu, line 292:
```suggestion
    """Run Trollstalker.
```

pnuu (approved):
LGTM. Naturally lots of corner cases that are not covered by the simple test, but at least the main functionality now has a test. One typo fix inline which I'll commit, and then merge this to `main`

PR #144, opened 2023-10-09
Update versioneer

mraspaud:
The old versioneer is not compatible with python 3.12, this PR updates versioneer


pnuu (approved):
LGTM

pnuu:
Seems that we have problems with botocore.

mraspaud:
yep :( looking into it...

mraspaud:
I think that's the culprit: https://github.com/aio-libs/aiobotocore/pull/1037

PR #152, opened 2024-02-08
Use factory function for publisher creation in trollstalker and switch to Watchdog

mraspaud:
This PR uses factory function for publisher creation in trollstalker and switches it to Watchdog


mraspaud:
The blacklisting of watchdog 4.0.0 is because of a bug solved in: https://github.com/gorakhargosh/watchdog/pull/1032

pnuu (approved):
LGTM

PR #154, opened 2024-04-23
Handle sensor list in topic

ninahakansson:
In case of list of sensor, use MULTIPLE_SENSORS in topic. We do not want a tuple or lsit in the publishing topic.

### DIFF
--- a/pytroll_collectors/scisys.py
+++ b/pytroll_collectors/scisys.py
@@ -62,8 +62,8 @@ JPSS_PLATFORM_NAME = {'npp': 'Suomi-NPP',
                       'jpss1': 'NOAA-20',
                       'noaa20': 'NOAA-20',
                       'noaa21': 'NOAA-21',
-                      'noaa21': 'NOAA-22',
-                      'noaa21': 'NOAA-23',
+                      'noaa22': 'NOAA-22',
+                      'noaa23': 'NOAA-23',
                       'j01': 'NOAA-20',
                       'j02': 'NOAA-21',
                       'j03': 'NOAA-22',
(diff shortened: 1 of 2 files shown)


pytroll_collectors/scisys.py
```diff
@@ -503,9 +503,19 @@ def stop(self):
         self.loop = False
 
 
+def update_sensor_tuple_or_list(to_send, config):
+    """Update sensor tuples and lists to MULTIPLE_SENSORS."""
+    to_send_dict = to_send.copy()
+    if "sensor" in to_send:
+        if type(to_send["sensor"]) is list or type(to_send["sensor"]) is tuple:
```
adybbroe, line 510: I think it is perhaps more pythonic to do `isinstance(to_send["sensor"], collections.abc.Sequence)`
?

pytroll_collectors/scisys.py
```diff
@@ -503,9 +503,19 @@ def stop(self):
...
+            to_send_dict["sensor"] = "MULTIPLE_SENSORS"
```
adybbroe, line 511: Should it be "multiple sensors" rather?

adybbroe (commented):
Apart from a small comment it looks good, thanks!

pytroll_collectors/scisys.py
mraspaud: I also prefer lower case

ninahakansson:
Unfortunately  isinstance("atms", collections.abc.Sequence) is also True ...

pytroll_collectors/scisys.py
gerritholl: `isinstance(to_send["sensor"], (list, tuple))` should also work

mraspaud (approved):
LGTM

PR #164, opened 2025-09-10
Switch to hatchling

mraspaud:
as a build system, drop versioneer, and update python versions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,10 +40,6 @@ jobs:
           python -m pip install --group test
           python -m pip install -e .[all]
 
-      - name: Test with pytest
-        run: |
-          pytest --cov=pytroll_watchers tests --cov-report=xml
-
       - name: Install unstable dependencies
         if: matrix.experimental == true
         shell: bash -l {0}


djhoese:
You need `hatch-vcs` too right?

pyproject.toml
```diff
@@ -0,0 +1,71 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "pytroll-collectors"
+dynamic = ["version"]
+description = "Pytroll data collectors"
+readme = "README.md"
+license = {file = "LICENSE"}
```
djhoese, line 10: Double check but I don't think this is the newest way to do this. For 1, you need to remove the license classifier in `classifiers`. Let me check what I did for other packages...

mraspaud: yes, I read the spec, and it's clear the sample project is outdated. Fixed in the latest commit.

.github/workflows/deploy_sdist.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v5
+
+      - name: Create sdist
+        shell: bash -l {0}
+        run: |
+          python -m pip install -q build
+          python -m build -s
```
djhoese, line 20: I had a coworker point this out, you should remove the `-s` so a wheel is also built. This makes installation faster for pure-python packages since `pip install` doesn't have to do it at install time.

pytroll_collectors/__init__.py
```diff
@@ -21,5 +21,3 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """The place where everything starts :)."""
-from . import _version
```
djhoese, line 24: This means there is no `pytroll_collectors.__version__` though right? I don't use this package but for consistency I'd like this re-added.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+license = "GPL-3.0-or-later"
+license-files = ["LICENSE"]
+requires-python = ">=3.11"
+authors = [
+    { name = "Martin Raspaud", email = "martin.raspaud@smhi.se" },
+]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
```
djhoese, line 19: Remove this line. Deprecated.

djhoese (changes requested):
Few small things. Plus:

You should change codecov `file:` to `files:` in `ci.yaml`. The `file` parameter is deprecated.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "pytroll-collectors"
+dynamic = ["version"]
+description = "Pytroll data collectors"
+readme = "README.md"
```
djhoese, line 9: Does pyproject.toml README definition handle the markdown versus restructuredtext determination for you? In `setup.py` you need to define `long_description_content_type`.

.github/workflows/deploy_sdist.yaml
```diff
@@ -0,0 +1,27 @@
+name: Deploy sdist
+
+on:
+  release:
+    types:
+      - published
```
djhoese, line 6: Do you want to enable this for pull requests too? The upload to PyPI looks like it is limited to release publishes so might be OK? Oh I guess pushes to `main` would be part of this too.

PR #168, opened 2025-09-30
Fix end time in segment gatherer

pnuu:
In some cases, like with VIIRS data, the `end_time` does not have the year in the file pattern and thus the date will be set as `1900-01-01` in Segment Gatherer. There is an existing fix in Geographic Gatherer that I here migrate to public method in `utils.py` and adjust a bit to make work also with segment gathering.

I also fixed the `datetime` imports in the files I touched.


mraspaud (approved):
LGTM, I just suggested the utcnow changes for the parts I thought were not problematic

PR #170, opened 2025-10-08
Switch to `netifaces-plus` and add a workaround to skip unusable network interfaces

pnuu:
The `netifaces` GitHub repository has been archived since 2021. This PR swithces to the somewhat more maintained `netifaces-plus` package that is a fork of the original and a drop-in replacement.

This change was actually triggered by the failure I got in running tests:
```python
>       inet_addrs = [netifaces.ifaddresses(iface).get(netifaces.AF_INET)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      for iface in netifaces.interfaces()]
E       ValueError: You must specify a valid interface name.
```
In my case this was triggered by `gpd0` interface created by GlobalProtect VPN.

I don't know what would be the best route to test this, mocking the `netifaces.interfaces()` to return "foo" doesn't really sound reasonable.

### DIFF
--- a/continuous_integration/environment.yaml
+++ b/continuous_integration/environment.yaml
@@ -21,7 +21,6 @@ dependencies:
   - freezegun
   - paramiko
   - responses
-  - netifaces
   - watchdog!=4.0.0
   - s3fs
   - pyinotify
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
LGTM, thanks for this! I think we have migrated posttroll, but I had forgotten about this one.

pnuu:
Yes, posttroll has been migrated.

Issue #171, opened 2025-10-10
Add support for logging configuration file in segment-gatherer


gerritholl:
Segment gatherer currently has a hardcoded logging configuration, logging everything to stderr or to a file with hardcoded internal log rotation:

https://github.com/pytroll/pytroll-collectors/blob/f062370ca571734085467d6d7c7f9346f144c5b9/pytroll_collectors/logging.py#L5-L27

Other pytroll packages, such as pytroll-watcher or trollflow2, can use an external logging configuration file `logging.yaml`.  It would be nice if the segment gatherer could also use `logging.yaml`.

pnuu:
Closed by #177

PR #177, opened 2026-04-06
Add `--log-config` support via yaml configuration file

adybbroe:
Add --log-config support to geographic_gatherer and segment_gatherer and initialize logging via shared setup_logging()

Preparing it also to be applicable for the trollstalker in the future should it be wished.

### DIFF
--- a/pytroll_collectors/tests/test_triggers.py
+++ b/pytroll_collectors/tests/test_triggers.py
@@ -25,11 +25,22 @@ class TestPostTrollTrigger:
         """Test timing out."""
         from pytroll_collectors.triggers import PostTrollTrigger
 
-        messages = [FakeMessage({"a": "a", 'start_time': 1, 'end_time': 2, 'collection_area_id': 'area_id',
+        time_tag_start = dt.datetime(2000, 1, 1, 12, 0)
+        one_min = dt.timedelta(seconds=60)
+        # messages = [FakeMessage({"a": "a", 'start_time': 1, 'end_time': 2, 'collection_area_id': 'area_id',
+        #                          'format': 'fmt', 'data_processing_level': 'l1b', 'uri': 'uri1'}),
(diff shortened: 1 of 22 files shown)


mraspaud:
Closes #171

pytroll_collectors/tests/test_triggers.py
```diff
@@ -25,11 +25,22 @@ def test_timeout(self, nssub, get_metadata):
...
 
-        messages = [FakeMessage({"a": "a", 'start_time': 1, 'end_time': 2, 'collection_area_id': 'area_id',
+        time_tag_start = dt.datetime(2000, 1, 1, 12, 0)
+        one_min = dt.timedelta(seconds=60)
+        # messages = [FakeMessage({"a": "a", 'start_time': 1, 'end_time': 2, 'collection_area_id': 'area_id',
+        #                          'format': 'fmt', 'data_processing_level': 'l1b', 'uri': 'uri1'}),
+        #             FakeMessage({"b": "b", 'start_time': 2, 'end_time': 3, 'collection_area_id': 'area_id',
+        #                          'format': 'fmt', 'data_processing_level': 'l1b', 'uri': 'uri2'}),
+        #             FakeMessage({"c": "c", 'start_time': 3, 'end_time': 4, 'collection_area_id': 'area_id',
+        #                          'format': 'fmt', 'data_processing_level': 'l1b', 'uri': 'uri3'})]
```
pnuu: This can be cleaned away.

pnuu (commented):
In general LTGM, one cleanup suggestion inline. As you said during the coffee break, the test coverage could be enhanced a bit.

Should the existing log-file option be deprecated with a warning message that the logging config should be used instead? What happens now if both are defined?

pytroll_collectors/tests/test_triggers.py
adybbroe: Done, thx

pnuu (approved):
LGTM, thanks!

PR #178, opened 2026-04-15
Start cat runner without requiring a nameserver

erlingmoeller-software:
Make it possible to disable subscriber nameserver in `cat.py`.


pnuu:
I made a small change: if subscriber nameser is set to `"false"` or `"False"` in the config, disable the nameserver. By default the old behaviour of nameserver on localhost is preserved.

PR #180, opened 2026-04-21
Allow fractional collection periods in Segment Gatherer

pnuu:
Currently Segment Gatherer is limited to collection periods for full minutes (e.g. 5/10/15 minutes). This PR adds the possibility to collect also to sub-minute periods, having the 2.5 minute repeat-cycle of MTG RSS in mind. The use of `datetime` objects came from AI, but the implementation it suggested didn't work (use of `%` modulo operator on `datetime` and `timedelta`), so I wrote the actual implementation myself.

I've also removed the deprecated use of `dt.datetime.utcnow()` from Segment Gatherer and its tests.

### DIFF
--- a/pytroll_collectors/tests/test_segments.py
+++ b/pytroll_collectors/tests/test_segments.py
@@ -250,7 +250,7 @@ class TestSegmentGatherer:
         fake_message = FakeMessage(mda)
         message = Message(fake_message, self.msg0deg._patterns['msg'])
         self.msg0deg._create_slot(message)
-        now = dt.datetime.utcnow()
+        now = dt.datetime.now(dt.timezone.utc)
         slot = self.msg0deg.slots[slot_str]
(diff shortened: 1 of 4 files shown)


pnuu:
CI workflow has been cancelled due to https://github.com/pytroll/posttroll/issues/86

pytroll_collectors/tests/test_segments.py
```diff
@@ -1529,7 +1565,6 @@ def test_remote_file_with_filesystem_passes_filesystem_info(filesystem):
         segment_gatherer.process(msg)
 
     timestamp = '2025-06-26 07:30:00'
-    print(segment_gatherer.slots)
```
mraspaud, line 1532: good catch!

mraspaud (approved):
LGTM, even if I don't like hairy caterpillars :)

PR #182, opened 2026-04-22
Fix the sigterm tests

mraspaud:
Python has become much more sensitive when forking, we need to take that into account when creating an external process for testing.

This in turn lead to neverending tests because the new posttroll authenticator thread (zmq) was never started.

### DIFF
--- a/pytroll_collectors/geographic_gatherer.py
+++ b/pytroll_collectors/geographic_gatherer.py
@@ -1,5 +1,6 @@
 """Geographic segment gathering."""
 
+from multiprocessing import Event
 import logging
 import signal
 import time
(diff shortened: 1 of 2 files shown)


pytroll_collectors/geographic_gatherer.py
```diff
@@ -33,7 +34,7 @@ def __init__(self, opts):
         self.triggers = []
         self.return_status = 0
 
-        self._sigterm_caught = False
+        self._sigterm_caught = Event()
```
pnuu, line 37: Is this for thread-safety or just more elegant way?

pnuu (approved):
LGTM, thanks for the fix(es)!

Just one question inline.

pytroll_collectors/geographic_gatherer.py
mraspaud: It's for thread-safety, even though that was not what hit us this time.

